PR #29, opened 2018-06-08
Add cors annotations and auth header configurable

Rajith90:
## Purpose
- Added CORS annotations to generated ballerina services
- Get authorization header from API , if not from the config file

## Goals
- Micro gateway apis should be able to restricted with CORS headers and , user should be able to specify any header to provide the access token

### DIFF
--- a/components/micro-gateway-core/src/main/ballerina/gateway/filters/authn_filter.bal
+++ b/components/micro-gateway-core/src/main/ballerina/gateway/filters/authn_filter.bal
@@ -52,13 +52,14 @@ public type AuthnFilter object {
         boolean isAuthorized;
         if (isSecured) {
             string authHeader;
-            if(request.hasHeader(AUTH_HEADER)) {
-                authHeader = request.getHeader(AUTH_HEADER);
+            string authHeaderName = getAuthorizationHeader(reflect:getServiceAnnotations(context.serviceType));
+            if (request.hasHeader(authHeaderName)) {
+                authHeader = request.getHeader(authHeaderName);
             } else {
                 log:printError("No authorization header was provided");
                 return createFilterResult(false, 401, "No authorization header was provided");
             }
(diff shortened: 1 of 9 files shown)


components/micro-gateway-core/src/main/ballerina/gateway/utils/utils.bal
```diff
@@ -346,3 +346,13 @@ public function setErrorMessageToFilterContext(http:FilterContext context, int e
     context.attributes[ERROR_MESSAGE] = errorMessage;
     context.attributes[ERROR_DESCRIPTION] = getFailureMessageDetailDescription(errorCode, errorMessage);
 }
+
+public function getAuthorizationHeader(reflect:annotationData[] annData) returns string {
+    APIConfiguration apiConfig = getAPIDetailsFromServiceAnnotation(annData);
+    string authHeader = apiConfig.authorizationHeader;
+    if( authHeader == "") {
```
harsha89: can we use StringUtiles here

components/micro-gateway-core/src/main/ballerina/gateway/filters/authn_filter.bal
```diff
@@ -52,13 +52,14 @@ public type AuthnFilter object {
         boolean isAuthorized;
         if (isSecured) {
             string authHeader;
-            if(request.hasHeader(AUTH_HEADER)) {
-                authHeader = request.getHeader(AUTH_HEADER);
+            string authHeaderName = getAuthorizationHeader(reflect:getServiceAnnotations(context.serviceType));
+            if(request.hasHeader(authHeaderName)) {
```
mushthaq33: Shall we format this?

components/micro-gateway-core/src/main/ballerina/gateway/utils/utils.bal
mushthaq33: Let's format as well.

PR #31, opened 2018-06-09
Add integration test framework for micro-gateway

Rajith90:
## Purpose
- Provide support to run micro gateway server and to perform integration tests while server is running

## Goals
- Achieve test coverage for micro-gateway


tests/pom.xml
```diff
@@ -0,0 +1,145 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
harsha89: add licence

PR #74, opened 2018-06-22
Adding capability to logging to an external file from CLI

malinthaprasan:
## Purpose
Adding capability to logging to an external file from CLI: `<CLI_HOME>/logs/micro-gw.log`

To configure logging, added `<CLI_HOME>/conf/logging.properties`

### DIFF
--- a/distribution/resources/cli-conf/logging.properties
+++ b/distribution/resources/cli-conf/logging.properties
@@ -18,11 +18,9 @@
 
 ####################### HANDLERS #######################
 # Configurations for console logging
-java.util.logging.ConsoleHandler.level=WARNING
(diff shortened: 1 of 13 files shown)


distribution/resources/cli-conf/logging.properties
```diff
@@ -0,0 +1,40 @@
+#
+# Copyright (c) 2017, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
```
mushthaq33: Shall we update the license?

ChamNDeSilva: Year of the license header need to be changed

malinthaprasan: Thanks. It is fixed now.

PR #98, opened 2018-06-28
Make authorization enable by default in gateway

Rajith90:


ruks:
Fixed https://github.com/wso2/product-microgateway/issues/115

PR #146, opened 2018-07-04
Adding a flag to enable/disable file uploading task.

dinusha92:
## Purpose
To enable or disable file uploading tasks in a distributed environment with shared file system.

## Goals
Adding a flag to define as true or false in conf.
Depending on the value, it will activate or deactivate the task.
By default it has been set to true.


dinusha92:
Fixing https://github.com/wso2/product-microgateway/issues/169

Issue #193, opened 2018-07-12
Configuration to disable SSL hostname verification for Key manager.


pubudu538:
**Description:**
api_gateway_endpoint.bal has the Key manager endpoint and it does not have verifyHostname configuration. This can be set as below.

endpoint http:Client keyValidationEndpoint {
    url:getConfigValue(KM_CONF_INSTANCE_ID, KM_SERVER_URL, "https://localhost:9443"),
    cache: { enabled: false },
    secureSocket:{
        verifyHostname:false
    }
};

We can come up with a config in our MG side to accommodate this.

**Suggested Labels:**

**Suggested Assignees:**

**Affected Product Version:**

**OS, DB, other environment details and versions:**    

**Steps to reproduce:**

**Related Issues:**

PR #209, opened 2018-07-25
Add configuration to disable SSL hostname verification for KM

msm1992:
## Purpose
This PR is to fix https://github.com/wso2/product-microgateway/issues/193


Rajith90:
Check whether we need to add host name verification skip for other endpoints as well

msm1992:
@Rajith90 It is only conditionRetrievalEndpoint other than KeyValidationEndpoint and it is not used anywhere in the code.

pubudu538:
Let's add this to the documentation.

PR #212, opened 2018-08-06
Combining request,response,execution events to a single stream

fazlan-nazeem:
## Purpose

With the move to SP in 2.6 we need to combine request,response,execution event streams into a single stream and publish it in the response path. This PR implements this functionality.

### DIFF
--- a/components/micro-gateway-core/src/main/ballerina/gateway/executiontime_util.bal
+++ b/components/micro-gateway-core/src/main/ballerina/gateway/executiontime_util.bal
@@ -18,33 +18,6 @@ import ballerina/log;
 import ballerina/io;
 import ballerina/http;
 
-
-function getExecutionTimePayload(ExecutionTimeDTO executionTimeDTO) returns string {
-    string output = executionTimeDTO.apiName + OBJ + executionTimeDTO.apiVersion + OBJ + executionTimeDTO.tenantDomain +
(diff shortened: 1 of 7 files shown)


components/micro-gateway-core/src/main/ballerina/gateway/executiontime_util.bal
```diff
@@ -58,20 +31,20 @@ function generateExecutionTimeEvent(http:FilterContext context) returns Executio
         executionTimeDTO.keyType = PRODUCTION_KEY_TYPE;
     }
     executionTimeDTO.apiName = getApiName(context);
-    executionTimeDTO.apiVersion = getAPIDetailsFromServiceAnnotation(reflect:getServiceAnnotations(context.serviceType)).
-    apiVersion;
+    executionTimeDTO.apiVersion = getAPIDetailsFromServiceAnnotation(reflect:getServiceAnnotations(context.serviceType))
```
chamilaadhi: Previous version we had the API version as <apiProvider>:v<apiversion> kind of a format (in synapse gateway event). Can you check whether it is the same here. If it is the same, can you remove it and get the version only. (same thing is done in synapse event. this would make both events similar)

fazlan-nazeem: In here the apiVersion will get values such as 1.0.0 etc. The provider name or prefix "v" is not added.

PR #284, opened 2019-01-23
Etcd 2.x

hisanhunais:
## Purpose

## Goals

## Approach

## User stories

etcd fallback scenarios

## Documentation
- Doc has impact

## Automation tests
 - Integration tests

## Test environment
-  JDK version - jdk1.8.0_144


components/micro-gateway-cli/src/main/java/org/wso2/apimgt/gateway/cli/cmd/SetupCmd.java
```diff
@@ -335,6 +354,9 @@ public void execute() {
                     apis.add(api);
                 }
             }
+//            for (ExtendedAPI api : apis) {
+//                outStream.println("ID for API " + api.getName() + " : " + api.getId());
+//            }
```
dinusha92: Can we remove the commented code segment.

components/micro-gateway-cli/src/main/resources/templates/prodEndpointReinit.mustache
```diff
@@ -0,0 +1,39 @@
+etcdKeyConfigValue = apiId + "_prod_etcdKey";
```
pubudu538: Can we think of a way to avoid duplicating code segments in this file as well as in the sandboxendpoint file? What if we have another common file where we pass the prod or sandbox endpoint to the common file.

components/micro-gateway-core/src/main/ballerina/gateway/endpoints/etcd_endpoint.bal
```diff
@@ -0,0 +1,29 @@
...
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+import ballerina/io;
+import ballerina/http;
+import ballerina/config;
+
+endpoint http:Client etcdEndpoint {
+    url: retrieveConfig("etcdurl", "http://127.0.0.1:2379"),
```
dinusha92: I think the URL mentioned here ( http://127.0.0.01:2379") would be subjected to changes. If that is the case, hard-coding such a value would not be the ideal solution here.

components/micro-gateway-core/src/main/ballerina/gateway/utils/etcd_util.bal
```diff
@@ -0,0 +1,258 @@
...
+@Description {value:"Periodic Etcd Query. Trigger function of etcd timer task"}
+public function etcdTimerTask() returns error? {
+    printDebug(KEY_ETCD_UTIL, "Etcd Periodic Query Initiated");
+    if(etcdUrls.count() > 0){
+        printDebug(KEY_ETCD_UTIL, "etcdurl map values - start");
+        foreach key, value in etcdUrls {
+            string currentUrl = <string>value;
+            string fetchedUrl = etcdLookup(<string>key);
+
+            if(currentUrl != fetchedUrl){
```
dinusha92: Please format the code.

components/micro-gateway-core/src/main/ballerina/gateway/utils/etcd_util.bal
```diff
@@ -0,0 +1,258 @@
...
+@Description {value:"Error function of etcd timer task"}
+public function etcdError(error e) {
+    printError(KEY_ETCD_UTIL, "Etcd Timer Task failed");
+}
+
+@Description {value:"Setting up etcd requirements"}
+public function etcdSetup(string key, string etcdConfigKey, string default) returns string {
+    string endpointUrl;
+
+    if(!etcdConnectionAttempted){
```
dinusha92: Formatting issues. Please check the whole code as there are multiple places.

tests/src/test/java/org/wso2/micro/gateway/tests/common/CLIExecutor.java
```diff
@@ -114,6 +114,43 @@ public void generateFromDefinition(String label, String project, String endpoint
...
+
+        System.setProperty(GatewayCliConstants.CLI_HOME, this.cliHome);
+        log.info("CLI Home: " + this.cliHome);
+
+        String config = new File(
+                getClass().getClassLoader().getResource("confs" + File.separator + "default-cli-test-config.toml")
+                        .getPath()).getAbsolutePath();
+        System.setProperty("user.dir", path.toString());
+        String[] args = { "setup", project, "--label", label, "--username", "admin", "--password",
+                "admin", "--server-url", "http://localhost:9443", "--truststore",
```
dinusha92: Shouldn't be the "localhost" configurable?

components/micro-gateway-core/src/main/ballerina/gateway/endpoints/etcd_endpoint.bal
```diff
@@ -0,0 +1,29 @@
+// Copyright (c)  WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
```
nimanthag: Year missing

nimanthag: Since all bal file does not contain this, It's better to keep the consistency.

tests/src/test/java/org/wso2/micro/gateway/tests/common/CLIExecutor.java
hisanhunais: This URL (http://localhost:9443) is the URL of the MockHttpServer used for test cases.

PR #290, opened 2019-01-29
Etcd Support Feature for microgateway master branch

hisanhunais:
## Purpose

## Goals

## Approach

## User stories

etcd fallback scenarios

## Documentation
- Doc has impact

## Automation tests
 - Integration tests

## Test environment
-  JDK version - jdk1.8.0_144


tests/src/test/java/org/wso2/micro/gateway/tests/context/Utils.java
```diff
@@ -267,5 +274,54 @@ public static void deleteFolder(File folder) {
...
+    }
+
+    /**
+     * Assert the result of a response
+     *
+     * @param response The response object.
+     * @param responseData The data which is expected as the response
+     * @param responseCode The response code which is expected
+     */
+    public static void assertResult(HttpResponse response, String responseData, int responseCode){
```
mushthaq33: Formatting issue.

tests/src/test/java/org/wso2/micro/gateway/tests/util/EtcdClient.java
```diff
@@ -0,0 +1,142 @@
...
+ * This class performs required function related to Etcd.
+ */
+public class EtcdClient {
+    private String etcdUrl;
+    private String EtcdKVBasePath = "/v3alpha/kv";
+    private String EtcdAuthBasePath = "/v3alpha/auth";
+    private String EtcdUserBasePath = EtcdAuthBasePath + "/user";
+    private String EtcdRoleBasePath = EtcdAuthBasePath + "/role";
+
+    public EtcdClient(String host, String port){
```
mushthaq33: Formatting issue.

components/micro-gateway-cli/src/main/resources/templates/endpointReinit.mustache
```diff
@@ -0,0 +1,13 @@
+etcdKey = gateway:retrieveConfig("{{api.name}}.{{api.version}}.{{endpointUrlType}}.etcd.key","{{api.name}}.{{api.version}}.{{endpointUrlType}}.endpoint.0");
+hasUrlChanged = <boolean>gateway:urlChanged[etcdKey];
+match hasUrlChanged{
+    boolean matchUrlChanged => reinitRequired=matchUrlChanged;
+    error err => io:println("error: " + err.message);
+}
+if(reinitRequired)
+{
+    io:println("Re-initializing the endpoint");
```
Rajith90: Lets use the log package instead of io

components/micro-gateway-cli/src/main/resources/templates/prodEndpointReinit.mustache
```diff
@@ -0,0 +1,39 @@
...
+                                gateway:urlChanged[etcdKey] = false;
+                            } catch (error err) {
+                                reinitFailed = true;
+                                gateway:urlChanged[etcdKey] = true;
+
+                                http:Response res = new;
+                                res.statusCode = 500;
+                                json payload = {
+                                    "fault": {
+                                        "code": "101505",
```
Rajith90: What is this error code

components/micro-gateway-core/src/main/ballerina/gateway/constants/constants.bal
```diff
@@ -343,6 +343,8 @@
 @final string KEY_OAUTH_PROVIDER = "OAuthAuthProvider";
 @final string KEY_UPLOAD_TASK = "UploadTimerTask";
 @final string KEY_ROTATE_TASK = "RotateTimerTask";
+@final string KEY_ETCD_UTIL = "EtcdUtil";
 
 
 @final public int DEFAULT_LISTENER_TIMEOUT = 120000; //2 mins
+@final public int DEFAULT_ETCD_TRIGGER_TIME = 10000; //10 seconds
```
Rajith90: Can we increase the default value, 10 seconds peiodic task seems too freqeunt. Is it fine to increase this value

hisanhunais: Yes we can increase this value. Any suggestions?
Also this value is overridable.

Rajith90: Lets make the default value 60 seconds

Rajith90:
This PR is replaced with : https://github.com/wso2/product-microgateway/pull/489

Issue #283, opened 2019-01-18
Add gRPC support for Micro-gateway 


Chamindu36:
**Description:**
gRPC is a high-performance, lightweight communication framework designed for making traditional RPC calls, and developed by Google. The framework uses HTTP/2, the latest network transport protocol, primarily designed for low latency and multiplexing requests over a single TCP connection using streams. This makes gRPC amazingly fast and flexible compared to REST over HTTP/1.1.
   Existing micro-gateway does not provide support for gRPC protocol. To include above mentioned benefits to micro-gateway it is required to implement this support into micro-gateway.

**Suggested Labels:**
Enhancement
type/feature

**Affected Product Version:**
2.6.0
3.0.0

**OS, DB, other environment details and versions:**    

**Steps to reproduce:**

**Suggested Approach:**
Implement gRPC support for product micro-gateway as a service which supports HTTP requests into gRPC supported ones. In this scenario we have to implement a ballerina service which receives REST requests and convert them to requests that can be identified in endpoint which is a gRPC service. But in this case we don't need to implement new gRPC filers to process request since receiving requests will be treated as REST/HTTP requests. After the filter processing these requests will be converted into gRPC supported ones. Then that converted ones will be sent to endpoint. When user provides the .proto API definition at Setup command level it reads the proto file and creates the ballerina file which act as service which do the REST to gRPC conversion. For that we have to implement logic to read .proto file from the scratch and implement the business logic of conversion according to the .proto file.

![qp2](https://user-images.githubusercontent.com/42435576/51366751-232a2480-1b0d-11e9-9786-726ef4f1cd47.png)

Figure 1

According to above approach this grpc feature is to be enabled at micro-gateway set up. At microgateway setup user required to provide .proto definition and the grpc endpoint. We will introduce a new setup command flag as “-oa” and “--openApi” which can be configured at setup command as “-oa=”PATH_TO_”.proto”_API_DEFINITION_FILE” to provide path of the “.proto” API definition file which required to setup micro-gateway with gRPC support. Then it requires to enter gRPC endpoint at the setup with “-e” flag. In this scenario user can only enter gRPC endpoint as the production endpoint. When this command is provided, the ballerina source will be generated in a way it can support above figure 1 diagram flaw.

[...]

PR #291, opened 2019-01-29
Add gRPC support for API product micro-gateway 2.x 

Chamindu36:
## Purpose
Existing micro-gateway does not provide support for gRPC protocol. To include above mentioned benefits to micro-gateway it is required to implement this support into micro-gateway.

## Goals
Add gRPC support for API micro-gateway.

## Approach
Approach is to Implement gRPC support for product micro-gateway as a service which supports HTTP requests into gRPC supported ones. In this scenario we have to implement a ballerina service which receives REST requests and convert them to requests that can be identified in endpoint which is a gRPC service. But in this case we don't need to implement new gRPC filers for process request since receiving requests will be treated as REST/HTTP requests. After the filter processing these requests will be converted into gRPC supported ones. Then that converted ones will be sent to endpoint. When user provides the .proto API definition at Setup command level it reads the proto file and creates the ballerina file which act as service which do the REST to gRPC conversion. For that we have to implement logic to read .proto file from the scratch and implement the business logic of conversion according to the .proto file.

image

Figure 1

According to above approach this grpc feature is to be enabled at micro-gateway set up. At microgateway setup user required to provide .proto definition and the grpc endpoint. We will introduce a new setup command flag as **“-oa”** and **“--openApi”** which can be configured at setup command as **“-oa=”PATH_TO_”.proto”_API_DEFINITION_FILE”** to provide path of the “.proto” API definition file which required to setup micro-gateway with gRPC support. Then it requires to enter gRPC endpoint at the setup with “-e” flag. In this scenario user can only enter gRPC endpoint as the production endpoint. When this command is provided, the ballerina source will be generated in a way it can support above figure 1 diagram flaw.

Then from the given path, cmd side of product micro-gateway which written in Java will extract methods, models and configurations from the file and generate the java objects related to that definition given in the .proto file. Ballerina grpc support has a skeleton which used to generate ballerina source that support gRPC protocol. Then java objects will be mapped into that skelton and generate “Client_stub.bal” file and generate “Client.bal” file which includes logic of converting HTTP request payloads into gRPC supported objects. And this Client.bal file has the ability to convert gRPC responses into HTTP/REST responses also. This ballerina source files has to be generated using mustache templates. 
In the mapping of HTTP/ REST to gRPC there is a optional field in the proto files which uses to convert Resource methods and it’s paths into their respective gRPC supported functions. After the resources are generated then using build command user can get an executable and start the server.

**This PR does not completely provide gRPC support for the API micro-gateway. Some of the functionalities have to be provided from Ballerina, and that support not exists yet.
For more reference  https://github.com/wso2/product-microgateway/issues/283**


components/micro-gateway-cli/src/main/java/org/wso2/apimgt/gateway/cli/codegen/CodeGenerator.java
```diff
@@ -33,12 +33,12 @@
...
 import org.wso2.apimgt.gateway.cli.model.template.service.ListenerEndpoint;
 import org.wso2.apimgt.gateway.cli.utils.CodegenUtils;
-import org.wso2.apimgt.gateway.cli.utils.OpenApiCodegenUtils;
 import org.wso2.apimgt.gateway.cli.utils.GatewayCmdUtils;
+import org.wso2.apimgt.gateway.cli.utils.OpenApiCodegenUtils;
 
 import java.io.File;
 import java.io.IOException;
-import java.nio.file.Paths;
+import java.nio.file.*;
```
npamudika: Lets do not use the wild card imports. Shall we fix this?

components/micro-gateway-cli/src/main/java/org/wso2/apimgt/gateway/cli/codegen/CodeGenerator.java
```diff
@@ -191,4 +189,42 @@ private String getContent(Object endpoints, String templateDir, String templateN
...
     }
+
+    /**
+     * Generates ballerina source for provided Open APIDetailedDTO Definition in {@code definitionPath}.
+     * Generated source will be written to a ballerina package at {@code outPath}
+     * <p>Method can be user for generating Ballerina mock services and clients</p>
+     *
+     * @throws IOException                  when file operations fail
+     * @throws BallerinaServiceGenException when code generator fails
+     */
```
npamudika: Shall we add the param definitions too?

components/micro-gateway-cli/src/main/java/org/wso2/apimgt/gateway/cli/utils/GatewayCmdUtils.java
```diff
@@ -34,13 +34,7 @@
...
 import org.wso2.apimgt.gateway.cli.model.rest.APICorsConfigurationDTO;
 
-import java.io.BufferedReader;
-import java.io.File;
-import java.io.FileInputStream;
-import java.io.FileWriter;
-import java.io.IOException;
-import java.io.InputStream;
-import java.io.InputStreamReader;
+import java.io.*;
```
npamudika: Lets not use the wild card imports.

components/micro-gateway-cli/src/main/java/org/wso2/apimgt/gateway/cli/utils/grpc/GRPCUtils.java
```diff
@@ -0,0 +1,267 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.wso2.apimgt.gateway.cli.constants.GatewayCliConstants;
+import org.wso2.apimgt.gateway.cli.exception.CLIInternalException;
+import org.wso2.apimgt.gateway.cli.utils.grpc.GrpcGen.BalFileGenerationUtils;
+import org.wso2.apimgt.gateway.cli.utils.grpc.GrpcGen.BalGenToolException;
+import org.wso2.apimgt.gateway.cli.utils.grpc.GrpcGen.DescriptorsGenerator;
+import org.wso2.apimgt.gateway.cli.utils.grpc.GrpcGen.OSDetector;
+
+import java.io.*;
```
npamudika: Same here

components/micro-gateway-cli/src/main/java/org/wso2/apimgt/gateway/cli/utils/grpc/GRPCUtils.java
```diff
@@ -0,0 +1,267 @@
...
+import java.net.URL;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Locale;
+
+import static org.wso2.apimgt.gateway.cli.utils.grpc.GrpcGen.BalFileGenerationUtils.*;
+import static org.wso2.apimgt.gateway.cli.utils.grpc.GrpcGen.BalGenerationConstants.*;
```
npamudika: Same here

components/micro-gateway-cli/src/main/java/org/wso2/apimgt/gateway/cli/utils/grpc/GrpcGen/BalFileGenerationUtils.java
```diff
@@ -0,0 +1,223 @@
...
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+package org.wso2.apimgt.gateway.cli.utils.grpc.GrpcGen;
+
+import com.google.protobuf.DescriptorProtos;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.*;
```
npamudika: Shall we fix all the wild card imports?

components/micro-gateway-cli/src/main/java/org/wso2/apimgt/gateway/cli/utils/grpc/GrpcGen/DescriptorsGenerator.java
```diff
@@ -0,0 +1,117 @@
+/*
+ *  Copyright (c) 2018, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
```
npamudika: Lets update the licence year

components/micro-gateway-cli/src/main/java/org/wso2/apimgt/gateway/cli/utils/grpc/GrpcGen/BalFileGenerationUtils.java
```diff
@@ -0,0 +1,230 @@
...
+     */
+    public static void saveFile(URL url, String file) {
+        try (InputStream in = url.openStream(); FileOutputStream fos = new FileOutputStream(new File(file))) {
+            int length;
+            byte[] buffer = new byte[1024]; // buffer for portion of data from
+            while ((length = in.read(buffer)) > -1) {
+                fos.write(buffer, 0, length);
+            }
+            fos.close();
+            in.close();
```
chamindias: Shouldn't this,
1. fos.close();
2. in.close();
go in the "finally" block?

components/micro-gateway-cli/src/main/java/org/wso2/apimgt/gateway/cli/utils/grpc/GrpcGen/DescriptorsGenerator.java
```diff
@@ -0,0 +1,122 @@
...
+import java.io.FileOutputStream;
+import java.io.FileInputStream;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Locale;
+
+import static org.wso2.apimgt.gateway.cli.utils.grpc.GrpcGen.BalFileGenerationUtils.*;
+import static org.wso2.apimgt.gateway.cli.utils.grpc.GrpcGen.BalGenerationConstants.*;
```
chamindias: We don't normally use wildcard imports, so shall we remove that and add only the necessary imports?

components/micro-gateway-cli/src/main/java/org/wso2/apimgt/gateway/cli/utils/grpc/GrpcGen/BalFileGenerationUtils.java
```diff
@@ -0,0 +1,230 @@
...
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.URL;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Locale;
+
+import static org.wso2.apimgt.gateway.cli.utils.grpc.GrpcGen.BalGenerationConstants.*;
```
npamudika: Shall we remove this wild card import too?

components/micro-gateway-cli/src/main/java/org/wso2/apimgt/gateway/cli/utils/grpc/GrpcGen/ProtocCommandBuilder.java
```diff
@@ -0,0 +1,49 @@
...
+ *  software distributed under the License is distributed on an
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+package org.wso2.apimgt.gateway.cli.utils.grpc.GrpcGen;
+
+
+import static org.wso2.apimgt.gateway.cli.utils.grpc.GrpcGen.BalGenerationConstants.*;
```
npamudika: Lets not use wild card imports.

components/micro-gateway-cli/src/main/java/org/wso2/apimgt/gateway/cli/utils/grpc/GrpcGen/BalFileGenerationUtils.java
tharindu1st: this was a try-resource block doesn't need to close it get auto close.
we can remove below lines

PR #306, opened 2019-02-28
Fix semantic issues with gateway ballerina files

NipunaMarcus:
## Purpose
Fix issues with analytics_request_filter.bal, mutualSSL_filter.bal, throttle_filter.bal

### DIFF
--- a/components/micro-gateway-core/src/main/ballerina/gateway/filters/throttle_filter.bal
+++ b/components/micro-gateway-core/src/main/ballerina/gateway/filters/throttle_filter.bal
@@ -21,171 +21,162 @@ import ballerina/config;
 import ballerina/time;
 
 public type ThrottleFilter object {
+    public map<string> deployedPolicies = {};
 
-    public map<string> deployedPolicies = new;
-
-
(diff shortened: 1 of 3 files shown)


components/micro-gateway-core/src/main/ballerina/gateway/filters/throttle_filter.bal
```diff
@@ -225,53 +220,62 @@ function isApplicationLevelThrottled(AuthenticationContext keyValidationDto) ret
...
     requestStreamDto.apiTenant = getTenantDomain(context);
     requestStreamDto.apiName = getApiName(context);
     requestStreamDto.appId = keyValidationDto.applicationId;
 
+    if (apiVersion is string) {
+        requestStreamDto.apiKey += ":" + apiVersion;
+        requestStreamDto.subscriptionKey +=":" + apiVersion;
+        requestStreamDto.resourceKey += "/" + apiVersion;
+    }
+
```
Rajith90: Why are we setting api version to these values

NipunaMarcus: Thats how the code was before. I handled the optional string value.

PR #308, opened 2019-02-28
Fix basicAuth and Constants for ballerina version 0.990.3 

NipunaRanasinghe:
## Purpose
$subject


components/micro-gateway-core/src/main/ballerina/gateway/basicAuth.bal
```diff
@@ -26,6 +26,10 @@ import ballerina/reflect;
 import ballerina/crypto;
 import ballerina/encoding;
 
+import constants;
+import dtos;
+import utils;
+
```
Rajith90: Do we need to import these, because all of these are in the same package. There is only one package, i.e.gateway

NipunaRanasinghe: It was added since module `gateway` was selected as the project source root folder, which is incorrect. Will fix it

PR #309, opened 2019-02-28
Fix validation filter issues

nadeeshaan:

### DIFF
--- a/components/micro-gateway-core/src/main/ballerina/gateway/filters/validation_filter.bal
+++ b/components/micro-gateway-core/src/main/ballerina/gateway/filters/validation_filter.bal
@@ -27,11 +27,11 @@ import ballerina/reflect;
 import ballerina/internal;
 
 public int errorItem = 0;
-public string requestPath;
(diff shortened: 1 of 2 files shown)


components/micro-gateway-core/src/main/ballerina/gateway/filters/validation_filter.bal
```diff
@@ -82,26 +81,29 @@ public type ValidationFilter object {
...
+                            //         //check whether there is a model available from the asigned model name
+                            //         if (models[modelName] != null) {
+                            //             model = models[modelName];
+                            //         }
+                            //     } else {
+                            //         //getting inline model
+                            //         model = k[SCHEMA];
+                            //     }
+                            // }
                         }
```
Rajith90: Why this part is commented out

nadeeshaan: This particular logic needs to be refactored analyzing the json structure

PR #311, opened 2019-03-02
Migrate subscription_filter.bal

rasika:

### DIFF
--- a/components/micro-gateway-core/src/main/ballerina/gateway/filters/subscription_filter.bal
+++ b/components/micro-gateway-core/src/main/ballerina/gateway/filters/subscription_filter.bal
@@ -28,7 +28,7 @@ public type SubscriptionFilter object {
                         returns boolean {
         int startingTime = getCurrentTime();
         checkOrSetMessageID(filterContext);
-        boolean result = doFilterRequest(caller, request, filterContext);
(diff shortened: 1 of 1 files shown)


components/micro-gateway-core/src/main/ballerina/gateway/filters/subscription_filter.bal
```diff
@@ -37,134 +37,131 @@ public type SubscriptionFilter object {
...
-                                printDebug(KEY_SUBSCRIPTION_FILTER, "Setting key type as " +
-                                        authenticationContext.keyType);
-                                runtime:getInvocationContext().attributes[KEY_TYPE_ATTR] = authenticationContext
-                                .keyType;
-                                filterContext.attributes[AUTHENTICATION_CONTEXT] = authenticationContext;
-                                printDebug(KEY_SUBSCRIPTION_FILTER, "Subscription validation success.");
-                                return true;
-                            }
-                        }
+        if (authScheme != AUTH_SCHEME_JWT){
```
Rajith90: This will continue to the logic for auth schemes which are not jwt. We need to skip this filter when auth scheme is not jwt

rasika: Good catch! Fixed it.

PR #339, opened 2019-03-25
Removing "Expect" header from the request

kavishkafernando:
## Purpose
Removing "Expect" header from the request if the request is expecting a 100 continue header.

### DIFF
--- a/components/micro-gateway-cli/src/main/resources/templates/service.mustache
+++ b/components/micro-gateway-cli/src/main/resources/templates/service.mustache
@@ -72,13 +72,7 @@ service {{cut qualifiedServiceName " "}} on {{#api.transport}}{{#equals . "http"
     }
     @gateway:RateLimit{policy : "{{resourceTier}}"}
     resource function {{operationId}} (http:Caller outboundEp, http:Request req) {
-    if (req.expects100Continue()) {
-        req.removeHeader("Expect");
-        var result = outboundEp->continue();
-        if (result is error) {
-        log:printError("Error sending response", err = result);
-        }
-    }
+        handleExpectHeaderFor{{cut qualifiedServiceName " "}}(outboundEp, req);
     {{#if api.isDefaultVersion}}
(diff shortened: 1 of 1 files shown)


components/micro-gateway-cli/src/main/resources/templates/service.mustache
```diff
@@ -72,7 +72,13 @@ service {{cut qualifiedServiceName " "}} on {{#api.transport}}{{#equals . "http"
     }
     @gateway:RateLimit{policy : "{{resourceTier}}"}
     resource function {{operationId}} (http:Caller outboundEp, http:Request req) {
-    gateway:checkExpectHeaderPresent(req);
+    if (req.expects100Continue()) {
+        req.removeHeader("Expect");
+        var result = outboundEp->continue();
+        if (result is error) {
+        log:printError("Error sending response", err = result);
```
Rajith90: Lets change this error message to add details like error while sending the continue response

components/micro-gateway-cli/src/main/resources/templates/service.mustache
```diff
@@ -72,7 +72,13 @@ service {{cut qualifiedServiceName " "}} on {{#api.transport}}{{#equals . "http"
...
+        }
+    }
```
Rajith90: Lets move this complete logic to a separate function in each service and generate function name like 
 handleExpectHeaderFor{{cut qualifiedServiceName " "}} and call that function in each resource.

components/micro-gateway-cli/src/main/resources/templates/service.mustache
kavishkafernando: fixed

components/micro-gateway-cli/src/main/resources/templates/service.mustache
kavishkafernando: fixed

PR #410, opened 2019-04-16
Initiate project contributing guide

praminda:
## Purpose
Introduce initial version of project contributing guideliness


praminda:
After completing the guide we can close #411

PR #423, opened 2019-04-18
Introduce init command

praminda:
## Purpose
Init command will replace the behavior of previous setup command. Generated project structure is also updated.
|_api_definitions
|_conf
|_definition.yaml
|_gen
|_interceptors
|_policies.yaml
|_target


praminda:
Fixes 325

Issue #454, opened 2019-04-24
OASAPI JWT token invocation test fails


hasuniea:
**Description:**
Open API JWT token invocation test case fails.

**Suggested Labels:**
3.0.0-alpha

**Suggested Assignees:**
@hasuniea 

**Affected Product Version:**
3.0.0-alpha

**Steps to reproduce:**
OASAP jwt test case

PR #455, opened 2019-04-24
Fixing OASAPI JWT token invocation test fails #454

hasuniea:

### DIFF
--- a/tests/src/test/resources/definition.yaml
+++ b/tests/src/test/resources/definition.yaml
@@ -5,18 +5,18 @@ apis:
     production_endpoint:
       type: 'load_balance'
       urls:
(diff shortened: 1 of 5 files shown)


praminda:
Fixes #454

tests/src/test/resources/definition.yaml
```diff
@@ -5,18 +5,18 @@ apis:
     production_endpoint:
       type: 'load_balance'
       urls:
-        - 'https://petstore.ballerina.io/'
-        - 'https://petstore.ballerina.io/'
+        - 'http://www.mocky.io/v2/5cbfdeb03100006600035ef1'
```
praminda: This will make the internet connection required for running tests I guess?

hasuniea: fixed it

PR #459, opened 2019-04-24
Create docker base image for micro-gw

Rajith90:


docker/Dockerfile
```diff
@@ -0,0 +1,5 @@
+
```
pubudu538: Let's add the license header

Issue #448, opened 2019-04-23
Moving extension filter and token_revocation_extension out of generated sources


Rajith90:
**Description:**
extension_filter.bal and token_revocation_extension.bal are given as extensions for API developers. So these files should be outside the generated sources and should be in the project root. 
Lets create a folder called extensions and add these files there. 

**Affected Product Version:**
3.0.0-alpha

PR #462, opened 2019-04-25
moving extension filter and token_revocation_extension out of generated sources

hasuniea:
Hence created a new folder called extensions and copied both extension_filter.bal and token_revocation_extension.bal files and refer the extension folder to the generated source.

### DIFF
--- a/components/micro-gateway-cli/src/main/java/org/wso2/apimgt/gateway/cli/cmd/Main.java
+++ b/components/micro-gateway-cli/src/main/java/org/wso2/apimgt/gateway/cli/cmd/Main.java
@@ -54,7 +54,7 @@ public class Main {
             logger.error(MICRO_GW + ": Error occurred while executing command.", e);
             Runtime.getRuntime().exit(1);
         } catch (CLIInternalException e) {
-            outStream.println(MICRO_GW + ":" + " + INTERNAL_ERROR_MESSAGE" + " - " + e.getMessage());
+            outStream.println(MICRO_GW + ":" + INTERNAL_ERROR_MESSAGE + " - " + e.getMessage());
             logger.error(e.getMessage(), e);
             Runtime.getRuntime().exit(1);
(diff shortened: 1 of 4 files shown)


praminda:
Fixes #448

components/micro-gateway-cli/src/main/java/org/wso2/apimgt/gateway/cli/cmd/Main.java
```diff
@@ -54,7 +54,7 @@ public static void main(String... args) {
             logger.error(MICRO_GW + ": Error occurred while executing command.", e);
             Runtime.getRuntime().exit(1);
         } catch (CLIInternalException e) {
-            outStream.println(MICRO_GW + ":" + " + INTERNAL_ERROR_MESSAGE" + " - " + e.getMessage());
+            outStream.println(MICRO_GW + ":" + INTERNAL_ERROR_MESSAGE + " - " + e.getMessage());
```
praminda: W can avoid creating a string for `:`
```suggestion
            outStream.println(MICRO_GW + ':' + INTERNAL_ERROR_MESSAGE + " - " + e.getMessage());
```

components/micro-gateway-cli/src/main/java/org/wso2/apimgt/gateway/cli/utils/GatewayCmdUtils.java
```diff
@@ -352,6 +355,28 @@ public static void createProjectStructure(String projectName) throws IOException
         File policyResFile = new File(policyResPath);
         File policesFile = new File(projectDir + File.separator + GatewayCliConstants.PROJECT_POLICIES_FILE);
 
+        String extensionResPath = getFiltersFolderLocation() + File.separator +
```
praminda: Is is possible to avoid two terms `filter`, `extension` and use a single term?

PR #470, opened 2019-04-27
Fixes #469

### LINKED ISSUE #469: Micro-gw build/import commands cannot be executed directly on copied project.
**Description:**
Immediately after extracting microgateway toolkit, micro-gw build/import commands does not work if the project folder is copied from some other place.

**Steps to reproduce:**
Extract micro-gw toolkit and set the environmental variables.
Copy an existing project, which is working properly and paste in your working directory.
execute micro-gw build <project_name>

VirajSalaka:

### DIFF
--- a/components/micro-gateway-cli/src/main/java/org/wso2/apimgt/gateway/cli/utils/ToolkitLibExtractionUtils.java
+++ b/components/micro-gateway-cli/src/main/java/org/wso2/apimgt/gateway/cli/utils/ToolkitLibExtractionUtils.java
@@ -1,0 +1,69 @@
+/*
+ * Copyright (c) 2019, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.wso2.apimgt.gateway.cli.utils;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.wso2.apimgt.gateway.cli.cmd.InitCmd;
+import org.wso2.apimgt.gateway.cli.constants.GatewayCliConstants;
+import org.wso2.apimgt.gateway.cli.exception.CLIInternalException;
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+
+public class ToolkitLibExtractionUtils {
+    private static final Logger LOGGER = LoggerFactory.getLogger(InitCmd.class);
+
+    /**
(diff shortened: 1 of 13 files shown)


components/micro-gateway-cli/src/main/java/org/wso2/apimgt/gateway/cli/utils/ToolkitLibExtractionUtils.java
```diff
@@ -0,0 +1,72 @@
...
+        } catch (IOException e) {
+            String message = "Error while unzipping platform and runtime while project setup";
+            LOGGER.error(message, e);
+            throw new CLIInternalException(message);
+        }
+    }
+
+    private static void extractBallerinaDist(String destination, String libPath, String baloPath, String breLibPath) throws IOException {
+        if (!Files.exists(Paths.get(destination))) {
+            ZipUtils.unzip(destination + GatewayCliConstants.EXTENSION_ZIP, destination, true);
```
Rajith90: Passing the value true here will add same jar twice to the class path. accepth the isAddtoClasspath as a method variable and pass it to ZipUtils.unzip method

components/micro-gateway-cli/src/main/java/org/wso2/apimgt/gateway/cli/utils/ToolkitLibExtractionUtils.java
```diff
@@ -0,0 +1,72 @@
...
+            String runtimeExtractedPath = libPath + File.separator + GatewayCliConstants.CLI_RUNTIME;
+            String platformExtractedPath =
+                    GatewayCmdUtils.getCLILibPath() + File.separator + GatewayCliConstants.CLI_PLATFORM;
+
+            extractBallerinaDist(platformExtractedPath, libPath, baloPath, breLibPath);
+            extractBallerinaDist(runtimeExtractedPath, libPath, baloPath, breLibPath);
+
+        } catch (IOException e) {
+            String message = "Error while unzipping platform and runtime while project setup";
+            LOGGER.error(message, e);
```
hasuniea: If you catch it and log it and rethrow it, then there's no way for the upstream code to know that you've already logged the exception, and so the same exception might get logged twice.

components/micro-gateway-cli/src/main/java/org/wso2/apimgt/gateway/cli/utils/ToolkitLibExtractionUtils.java
VirajSalaka: fixed.

components/micro-gateway-cli/src/main/java/org/wso2/apimgt/gateway/cli/utils/ToolkitLibExtractionUtils.java
VirajSalaka: Removed the logger message.

Rajith90: Please check before whether upstream method logs it. Its better to log it twice rather than not logging it

VirajSalaka: Whenever CLIInternalException or CLIRuntimeException is thrown, it is logged in the Main Class. So it would be better to remove that additional log message.

hasuniea: I checked the code. Upstream(Main class) logs the error message.

praminda:
Fixes #469

Issue #487, opened 2019-05-02
Missing debug param from init command


hasuniea:
**Description:**
Missing debug param from init command.
**Suggested Labels:**

**Suggested Assignees:**

**Affected Product Version:**

Issue #486, opened 2019-05-02
Analytics truststore issue


hasuniea:
**Description:**
Ballerina truststore doesn't contain the analytics public cert. This issue is occurred due to the analytics server has updated the trust store itself.
**Suggested Labels:**

**Suggested Assignees:**

**Affected Product Version:**

**OS, DB, other environment details and versions:**    

**Steps to reproduce:**
1. enable the analytics from the micro-gw.conf
1. bash gateway

PR #488, opened 2019-05-02
 fixes for #486 and #487

hasuniea:


praminda:
@hasuniea Since we removed runtime from distribution with #490, fix done for #486 in this PR should be updated now. Fix for #487 is fine. Can you please update the PR with the changes

praminda:
Replaced with updated PR #538

PR #489, opened 2019-05-02
Etcd Support Feature for microgateway 3.x

hisanhunais:
## Purpose

## Goals

## Approach

## User stories

etcd fallback scenarios

## Documentation
- Doc has impact

## Automation tests
 - Integration tests

## Test environment
-  JDK version - jdk1.8.0_144
- Etcd version - etcd v3.3.9


Rajith90:
Temporarily closing the PR, until tested with the 4.x branch

PR #493, opened 2019-05-07
Dev-First Approach using openAPI definition only. Fixes #325

### LINKED ISSUE #325: Restructuring the Microgateway Project Folder Structure and Command Execution to Support Multiple APIs
Description:
There is a requirement to add the following features to the Microgateway in Future. 
Add multiple APIs to a project
Publish APIs from Microgateway to Store
Support fine-grained endpoint configurations based on resource level.

- Ex. GET menu - <IP_1>
- POST order - <IP_2>

Currently, when the “micro-gw setup” command is executed, it fails if the project already exists and therefore we have no mechanism to add another API to the same project. 

Considering the developer first approach, the current folder structure of the project is as follows.

Command: 

- **micro-gw setup <project_name> -oa <path_to_swagger_file> -e <endpoint>**

Folder Structure:
Screen Shot 2019-03-12 at 1 52 13 PM

**Proposed Flow of the Process:**

_setup command_

- **micro-gw setup <project_name>**

This is allowed and it will create the following folder structure
Screen Shot 2019-03-12 at 1 54 24 PM

- **micro-gw setup <project_name> -oa <path_to_swagger_file> -e <_endpoint>**

Folder structure will look like this,

Screen Shot 2019-03-12 at 1 54 31 PM

As you can see there is another folder added API_files. We will read the swagger and name accordingly. The version number is identified by reading the swagger file. 

 The purpose behind saving the swagger is to publish the API to the store from microgateway. Metadata file holds the metadata related to API. Other than that the endpoint will be added to the routes.yaml with the corresponding api_name and the version number.  This endpoint is the basic endpoint for the API. ( compare the difference with add-route command)

	
More importantly, there will not be a code generation operation at the “setup” phase. In other words, there won’t be a “pizzashackAPI_1.0.0.bal” file during this phase. 

			
_Set-project command_ 
	
- **micro-gw set-project <project-name>**

Along with this new structural implementation, there will be a lot of additional commands related to the project (mentioned in below sections). The developer will have to mention the project name within each and every command. This command removes that burden from the developer.

_add-api command_

As mentioned earlier, A project can either be “setup” with just one API definition or without any API definition. To add more APIs, we need to introduce the add command.

- **micro-gw add-api  -oa <path_to_swagger_file> -e endpoint**

As in the setup command, we will be continuing with api_name provided in the swagger.

consider the following command, (in this case API name is DrinksAPI
			micro-gw add-api  -oa drinks_swagger.json -e some_endpoint
	

Screen Shot 2019-03-12 at 1 54 55 PM

It can be observed that still there is no code generation (bal file generation) has happened. 

_Add-route command_
	

- **micro-gw add-route -a <user_provided_api_name> -v  <_version>  -m <http_method> -r <resource_name> -e <_endpoint>**

We have added a basic endpoint while setting up the API. This command is to provide separate endpoints based on resource level. <http_method> would be something like “GET”, “POST” etc. and resource_name would be something like “menu”. These details would be added to the routes.yaml as well. 

_Build command_

- **micro-gw build <project_name>**

This is the same existing command. The difference is that code generation and compilation both happen at this phase. (Previously code generation happened at the “setup”  phase.) In this case, pizzashackAPI_1.0.0.bal file and drinksAPI_1.0.0.bal both will be saved in the src directory. Target directory will have all the components which will be required to run the microgateway. 

The folder structure will look like this,
Screen Shot 2019-03-12 at 1 55 12 PM

_delete_api command_

- **micro-gw delete-api -a <API_name> -v <API_version>**

	Since we are supporting add_api command, we need to support delete_api command as well. 

_list_apis command_
	

- **micro-gw list_apis** 

	This will list down all the APIs with their available versions. (API name would be the name of that folder corresponding to the API) 

_register command_

- **micro-gw register -a <API_name> -v <API_version>** 

	This command will publish the API in Store. Since we have the swagger file saved in the directory, we can use the Publisher API.

The command names and the arguments are not finalized yet. But this is the expected way to accomplish the task. 

The structure of routes.yaml file looks like this,
Screen Shot 2019-03-12 at 12 43 02 PM

User can define the endpoint for the API in three different levels,

- API level
- Resource level w/o HTTP-method
- Resource level with HTTP-method

Further this routes file has capability to store any mediation associated with a certain resource.

Suggested Labels:
Docs has Impact, Feature

VirajSalaka:

### DIFF
--- a/components/micro-gateway-cli/src/main/java/org/wso2/apimgt/gateway/cli/cmd/BuildCmd.java
+++ b/components/micro-gateway-cli/src/main/java/org/wso2/apimgt/gateway/cli/cmd/BuildCmd.java
@@ -117,7 +117,6 @@ public class BuildCmd implements GatewayLauncherCmd {
                 etcd.setEtcdEnabled(GatewayCmdUtils.getEtcdEnabled(projectName));
                 GatewayCmdUtils.setEtcd(etcd);
 
-                //MgwDefinitionBuilder.build(projectName);
                 CodeGenerator codeGenerator = new CodeGenerator();
                 ThrottlePolicyGenerator policyGenerator = new ThrottlePolicyGenerator();
 
@@ -127,9 +126,6 @@ public class BuildCmd implements GatewayLauncherCmd {
                         GatewayCmdUtils.getProjectGenSrcInterceptorsDirectoryPath(projectName));
                 codeGenerator.generate(projectName, true);
 
-                //to indicate the api information which is not used in the code generation process, but included in
-                //definition.yaml
-                //MgwDefinitionBuilder.FindUnusedAPIInformation();
                 //Initializing the ballerina project and creating .bal folder.
                 InitHandler.initialize(Paths.get(GatewayCmdUtils.getProjectGenDirectoryPath(projectName)), null,
                         new ArrayList<>(), null);
--- a/components/micro-gateway-cli/src/main/java/org/wso2/apimgt/gateway/cli/cmd/BuildCmd.java
+++ b/components/micro-gateway-cli/src/main/java/org/wso2/apimgt/gateway/cli/cmd/BuildCmd.java
@@ -117,7 +117,7 @@ public class BuildCmd implements GatewayLauncherCmd {
                 etcd.setEtcdEnabled(GatewayCmdUtils.getEtcdEnabled(projectName));
                 GatewayCmdUtils.setEtcd(etcd);
 
-                MgwDefinitionBuilder.build(projectName);
+                //MgwDefinitionBuilder.build(projectName);
                 CodeGenerator codeGenerator = new CodeGenerator();
                 ThrottlePolicyGenerator policyGenerator = new ThrottlePolicyGenerator();
 
@@ -129,7 +129,7 @@ public class BuildCmd implements GatewayLauncherCmd {
 
                 //to indicate the api information which is not used in the code generation process, but included in
                 //definition.yaml
-                MgwDefinitionBuilder.FindUnusedAPIInformation();
+                //MgwDefinitionBuilder.FindUnusedAPIInformation();
                 //Initializing the ballerina project and creating .bal folder.
                 InitHandler.initialize(Paths.get(GatewayCmdUtils.getProjectGenDirectoryPath(projectName)), null,
                         new ArrayList<>(), null);
--- a/components/micro-gateway-cli/src/main/java/org/wso2/apimgt/gateway/cli/model/template/service/BallerinaOperation.java
+++ b/components/micro-gateway-cli/src/main/java/org/wso2/apimgt/gateway/cli/model/template/service/BallerinaOperation.java
@@ -19,6 +19,7 @@ package org.wso2.apimgt.gateway.cli.model.template.service;
 import io.swagger.v3.oas.models.ExternalDocumentation;
 import io.swagger.v3.oas.models.Operation;
 import io.swagger.v3.oas.models.parameters.Parameter;
+import org.wso2.apimgt.gateway.cli.constants.OpenAPIConstants;
 import org.wso2.apimgt.gateway.cli.exception.BallerinaServiceGenException;
 import org.wso2.apimgt.gateway.cli.model.mgwcodegen.MgwEndpointConfigDTO;
 import org.wso2.apimgt.gateway.cli.model.rest.ext.ExtendedAPI;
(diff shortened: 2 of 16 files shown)


components/micro-gateway-cli/src/main/java/org/wso2/apimgt/gateway/cli/cmd/BuildCmd.java
```diff
@@ -117,7 +117,7 @@ public void execute() {
                 etcd.setEtcdEnabled(GatewayCmdUtils.getEtcdEnabled(projectName));
                 GatewayCmdUtils.setEtcd(etcd);
 
-                MgwDefinitionBuilder.build(projectName);
+                //MgwDefinitionBuilder.build(projectName);
```
Rajith90: Lets remove the commented out codes

components/micro-gateway-cli/src/main/java/org/wso2/apimgt/gateway/cli/cmd/BuildCmd.java
```diff
@@ -129,7 +129,7 @@ public void execute() {
 
                 //to indicate the api information which is not used in the code generation process, but included in
                 //definition.yaml
-                MgwDefinitionBuilder.FindUnusedAPIInformation();
+                //MgwDefinitionBuilder.FindUnusedAPIInformation();
```
Rajith90: Lets remove the commented out codes

components/micro-gateway-cli/src/main/java/org/wso2/apimgt/gateway/cli/utils/MgwDefinitionBuilder.java
```diff
@@ -49,38 +49,36 @@
...
+//        }
+//
+//        try {
+//            //update the interceptor map
+//            setInterceptors();
+//        } catch (IOException e) {
+//            throw new CLIRuntimeException("Error while reading the '" + GatewayCliConstants.PROJECT_INTERCEPTORS_DIR +
+//                    "' directory");
+//        }
+//    }
```
Rajith90: Lets remove the commented out codes

components/micro-gateway-cli/src/main/java/org/wso2/apimgt/gateway/cli/utils/MgwDefinitionBuilder.java
```diff
@@ -140,7 +138,7 @@ public static String getRequestInterceptor(String basePath, String path, String
         }
         String interceptor = definitionConfig.getApis().getApiFromBasepath(basePath).getPathsDefinition().getMgwResource(path).
                 getEndpointListDefinition(operation).getRequestInterceptor();
-        validateInterceptorAvailability(requestInterceptorMap, interceptor, basePath, path, operation);
+        //validateInterceptorAvailability(requestInterceptorMap, interceptor, basePath, path, operation);
```
Rajith90: Lets remove the commented out codes

VirajSalaka: Already fixed via a PR added later.

components/micro-gateway-cli/src/main/java/org/wso2/apimgt/gateway/cli/utils/MgwDefinitionBuilder.java
VirajSalaka: Already fixed via a PR added later.

components/micro-gateway-cli/src/main/java/org/wso2/apimgt/gateway/cli/cmd/BuildCmd.java
VirajSalaka: Already fixed via a PR added later.

components/micro-gateway-cli/src/main/java/org/wso2/apimgt/gateway/cli/cmd/BuildCmd.java
VirajSalaka: Already fixed via a PR added later.

components/micro-gateway-cli/src/main/java/org/wso2/apimgt/gateway/cli/model/template/service/BallerinaOperation.java
```diff
@@ -76,18 +78,31 @@ public BallerinaOperation buildContext(Operation operation, ExtendedAPI api) thr
         this.externalDocs = operation.getExternalDocs();
         this.parameters = new ArrayList<>();
         this.methods = null;
-        Map<String, Object> extension =  operation.getExtensions();
-        if(extension != null){
-            Optional<Object> resourceTier = Optional.ofNullable(extension.get(X_THROTTLING_TIER));
+        Map<String, Object> extensions =  operation.getExtensions();
+        if(extensions != null){
```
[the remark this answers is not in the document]
VirajSalaka: Already fixed via a PR added later.

PR #498, opened 2019-05-08
Update README with new swagger extension samples

Rajith90:

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,8 +37,10 @@ Running the WSO2 API Microgateway is a 3 step process. The first two steps are i
 ##### Initializing a WSO2 API Microgateway project
 
 Initializing a WSO2 API Microgateway project creates the default folder structure at the location where the command is run.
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -237,46 +253,29 @@ public function validateResponse (http:Caller outboundEp, http:Response res) {
...
+x-mgw-production-endpoints:
+  urls:
+  - https://petstore.swagger.io/v2
 ```
 
-#### 3. Add resource level throttling policies
-API developer can specify the rate limiting policies for each resource. These policies should be defined in the policies.yaml file in the project directory
-By default set of policies are available, but user can add more policies to the file and later refer them by name in the definition.yaml
-
+Sample open API definition for interceptors can be found [here](samples/interceptors_sample.yaml).
```
VirajSalaka: IMO, It would be better to mention the structure of the interceptor ballerina function.

Rajith90: There are samples of interceptor functions in the README file

Issue #567, opened 2019-05-27
Set sslVerifyClient property to "optional" as default in micro-gw.conf


hasuniea:
**Description:**
If someone set the **sslVerifyClient** property to "optional", Authentication will be changed as per the request type. If a user invokes an API with OAuth header, The particular API is secured using Oauth token unless it will be secured using mutual SSL. Ultimately, users will be given the flexibility to select either OAuth2 token or mutual SSL or both as their API security option.
If someone set the **sslVerifyClient** property to "require" API must be secured by mutual ssl.

**Suggested Labels:**
MC-GW 3.0.0 RC1

**Affected Product Version:**
MC-GW 3.0.0 RC1

PR #569, opened 2019-05-27
setting the sslVerifyClient property to "optional" as default in micro-gw.conf

hasuniea:
## Purpose
If someone set the sslVerifyClient property to "optional", Authentication mechanism will be changed as per the request type. If a user invokes an API with OAuth header, The particular API needs to be secured using OAuth token unless it will be secured using mutual SSL. Ultimately, users will be given the flexibility to select either OAuth2 token or mutual SSL or both as their API security option.

If someone set the sslVerifyClient property to "require" API must be secured by mutual SSL.


praminda:
Fixes #567

PR #581, opened 2019-05-30
Fix #580: Change Open API extensions

Rajith90:

PR #589, opened 2019-06-03
removing swagger reference from validator

hasuniea:


praminda:
Related to schema validation fix done in #581

PR #590, opened 2019-06-03
Fix #556 : Custom auth header in open API

### LINKED ISSUE #556: Open API approach - Support define custom authorization header per API
**Description:**
When the open API approach is used in micro gateway we can not specify authorization header for the API. This can be changed only when APIs are imported from WSO2 API Manager.
For the APIs exposed using open API we have to use the default "Authorization" header to send the token/ auth credentials. Other wise will have to change  the header name in micro-gw.conf file, which will globally apply for all the APIs

Rajith90:


components/micro-gateway-core/src/main/ballerina/gateway/constants/api_security_constants.bal
```diff
@@ -141,14 +141,15 @@ public function getAuthenticationFailureMessage(int errorCode) returns string {
...
+public function getFailureMessageDetailDescription(http:FilterContext context, int errorCode, string errorMessage) returns string {
     string errorDescription = errorMessage;
     if (API_AUTH_INCORRECT_API_RESOURCE == errorCode) {
         errorDescription += DESCRIPTION_SEPARATOR + API_AUTH_INCORRECT_API_RESOURCE_DESCRIPTION;
     } else if (API_AUTH_INCORRECT_API_RESOURCE == errorCode) {
         errorDescription += DESCRIPTION_SEPARATOR + API_AUTH_ACCESS_TOKEN_INACTIVE_DESCRIPTION;
     } else if (API_AUTH_MISSING_CREDENTIALS == errorCode) {
-        errorDescription += DESCRIPTION_SEPARATOR + API_AUTH_MISSING_CREDENTIALS_DESCRIPTION;
+        string authHeaderName = getAuthorizationHeader(serviceAnnotationMap[getServiceName(context.serviceName)] ?: []);
+        errorDescription += DESCRIPTION_SEPARATOR + "Make sure your API invocation call has a header: \"" + authHeaderName + "\"";
```
praminda: Avoid string initialization for a single character
```suggestion
        errorDescription += DESCRIPTION_SEPARATOR + "Make sure your API invocation call has a header: \"" + authHeaderName + '"';
```

components/micro-gateway-cli/src/main/java/org/wso2/apimgt/gateway/cli/utils/OpenAPICodegenUtils.java
```diff
@@ -206,7 +206,11 @@ public static String getOpenAPIAsJson(OpenAPI openAPI, String openAPIContent, Pa
         }
         if (api.getCorsConfiguration() != null) {
             extensionsMap.put(OpenAPIConstants.CORS, api.getCorsConfiguration());
-        } return extensionsMap;
+        }
+        if(api.getAuthorizationHeader() != null) {
```
praminda:
```suggestion
        if (api.getAuthorizationHeader() != null) {
```

components/micro-gateway-cli/src/main/java/org/wso2/apimgt/gateway/cli/utils/OpenAPICodegenUtils.java
```diff
@@ -458,6 +462,11 @@ public static void setAdditionalConfigsDevFirst(ExtendedAPI api, OpenAPI openAPI
                 // explicitly set the cors enabled value to true if cors config found in the open API definition
                 api.getCorsConfiguration().setCorsConfigurationEnabled(true);
             }
+            // set authorization header from the open API extension
+            if (openAPI.getExtensions().get(OpenAPIConstants.AUTHORIZATION_HEADER) != null) {
```
praminda: How about ...
```java
Object authHeader = openAPI.getExtensions().get(OpenAPIConstants.AUTHORIZATION_HEADER);
if (authHeader != null) {
    api.setAuthorizationHeader(authHeader.toString());
}
```

components/micro-gateway-core/src/main/ballerina/gateway/constants/api_security_constants.bal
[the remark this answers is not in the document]
Rajith90: Since this is ballerina lets strict with double quotation

Rajith90: ince this is ballerina lets strict with double quotation

PR #663, opened 2019-06-26
fix the issue of running java command twice during build command

### LINKED ISSUE #660: In Windows, the microgateway is built successfully but an error is shown.
### Description:
In windows, the following log appears when build command is issued. But the project is built successfully. 
>>Compiling source
    src:0.0.0
>>
>>Running tests
    src:0.0.0
        No tests found
>>
>>Generating executable
    ./target/C:\Users\wso2\Documents\product-microgateway\tests\target\userProject3240826263572306658\resourceEndpointProject\target\resourceEndpointProject.balx
[Only one main parameter allowed but found several: "resourceEndpointProject" and "--compiled"]

This is due to the fact that when the build command is issued, java cli is executed twice and the second time it is executed with additional parameter "--compiled".

### Steps to reproduce:
Init microgateway project and then create/copy an openAPI definition to the api-definitions directory. Then execute the build command.

### Affected Product Version:
3.0.1

- OS: Windows 10

VirajSalaka:
### Purpose
Fix the issue of running java command twice during the build command when executed in windows.

### Issues
Fixes #660

### Automation tests
 - Unit tests added: No
 - Integration tests added: No

### Tested environments
Windows 10

---
- [x] Assigned 'Type' label
- [x] Assigned the project

### DIFF
--- a/distribution/resources/bin/micro-gw.bat
+++ b/distribution/resources/bin/micro-gw.bat
@@ -134,7 +134,7 @@ goto :end
                 if EXIST "%TARGET_DIR%\*.balx"  DEL /F "%TARGET_DIR%\*.balx"
                 call ballerina build src -o %TARGET_DIR%\%project_name:\=%.balx --offline --experimental --siddhiruntime
             POPD
-goto end
+goto :end
 
 :commandDebug
 	if %verbose%==T ECHO Running commandDebug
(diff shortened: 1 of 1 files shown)


distribution/resources/bin/micro-gw.bat
```diff
@@ -134,17 +134,7 @@ goto :end
...
-            SET originalArgs=%originalArgs% --compiled
-
-            REM Check for a debug param by looping through the remaining args list
-            :checkDebug
-                SHIFT
-                if ""%1""=="""" goto passToJar
-                if ""%1""==""--java.debug""  goto commandDebug
-            goto checkDebug
-goto :passToJar
+goto end
```
praminda:
```suggestion
goto :end
```

PR #712, opened 2019-07-25
Fix #635 - Multiple scope validation issue

### LINKED ISSUE #635: Scope validation fails if multiple scopes are defined for the resource
### Description:
If we add multiple scopes to same resource in the open API definition, API invocation fails even with correct jwt as well.

### Steps to reproduce:
1. Add the following security schema to any resource
`security:
        - petstore_auth:
            - write:pets
            - read:pets`

2. The invoke the resource with jwt having the scope "write:pets"
3. This scenario should works fine. But it gives error code 900910, which is scope validation error code
### Affected Product Version:
3.0.1

- OS: MacOs

---
#### Suggested Assignees:
@Rajith90

Rajith90:
### Purpose
If we add multiple scopes to same resource in the open API definition, API invocation fails even with correct jwt as well.
This PR fixes that issue

### Issues

Fixes #635 

### Automation tests
 - Unit tests added: No
 - Integration tests added: Will provide with test refactoring

### Tested environments

Not Tested

---
- [x] Assigned 'Type' label
- [x] Assigned the project

### DIFF
--- a/components/micro-gateway-cli/src/main/java/org/wso2/apimgt/gateway/cli/utils/OpenAPICodegenUtils.java
+++ b/components/micro-gateway-cli/src/main/java/org/wso2/apimgt/gateway/cli/utils/OpenAPICodegenUtils.java
@@ -824,6 +824,7 @@ public class OpenAPICodegenUtils {
             }
             //generate scopes string
             for (String scope : scopeList) {
+                scope = "\"" + scope + "\"";
                 scopes = StringUtils.isEmpty(scopes) ? scope : scopes + "," + scope;
             }
         }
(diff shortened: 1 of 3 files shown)


components/micro-gateway-cli/src/main/java/org/wso2/apimgt/gateway/cli/utils/OpenAPICodegenUtils.java
```diff
@@ -824,6 +824,7 @@ private static void setMgwAPISecurityAndScopes(ExtendedAPI api, OpenAPI openAPI)
             }
             //generate scopes string
             for (String scope : scopeList) {
+                scope = "\"" + scope + "\"";
```
praminda:
```suggestion
                scope = '"' + scope + '"';
```
Will work here? we can avoid additional string initialization.

PR #813, opened 2019-09-25
Fix k8s labels, secrets and annotations

### LINKED ISSUE #734: Cannot add labels to deployment using deployment toml
### Description:
I tried to define a label for the deployment in K8s and it failed. This feature needs to test properly along with other annotations for K8s.

[kubernetes]
  [kubernetes.kubernetesDeployment]
    enable = true
    image = 'pubudu/petstore:v1.0'
    labels = '{app:k8s_project}'

Sample definition in Ballerina for label defining https://github.com/ballerinax/kubernetes/blob/85411c2e3d76425bc825391267f6b8d9be72f7b6/kubernetes-extension-test/src/test/resources/job/ballerina_job.bal#L25

- OS:
- Client:
- DB:
- Env (Docker/K8s):

---

praminda:
# Purpose

Default configuration model provided for k8s Secrets was incorrect. Secret configs should be given as a table.
Also label and annotation configs contain quotes which should be escaped before building the source file.
### Issues

Fixes #734 

### Automation tests
 - Unit tests added: No
 - Integration tests added: No

### Tested environments

MacOS

---
- [x] Assigned 'Type' label
- [x] Assigned the project

### DIFF
--- a/components/micro-gateway-cli/src/main/resources/default-deployment-config.toml
+++ b/components/micro-gateway-cli/src/main/resources/default-deployment-config.toml
@@ -23,7 +23,7 @@
   [kubernetes.kubernetesDeployment]
     enable = false
     #name = ''
-    #labels = ''
+    #labels = '{"", ""}'
     #replicas = ''
     #enableLiveness = ''
     #initialDelaySeconds = ''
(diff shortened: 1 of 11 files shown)


components/micro-gateway-cli/src/main/resources/default-deployment-config.toml
```diff
@@ -98,18 +98,19 @@
...
     #[[kubernetes.kubernetesConfigMap.configMaps]]
-    #    name = ''
-    #    mountPath = ''
-    #    readOnly = false
-    #    data = ['']
+      #name = ''
+      #mountPath = ''
+      #data = ['']
+      #readOnly = false
   [kubernetes.kubernetesPersistentVolumeClaim]
```
Rajith90: Lets add a label example to the default config

praminda: Fixed with 43012f6

PR #828, opened 2019-10-02
Download and add latest in-place tool

nihlaakram:
### Purpose
Download and add latest in-place tool to distribution. For platform specific distributions, the appropriate in-place executable is added.

### DIFF
--- a/distribution/pom.xml
+++ b/distribution/pom.xml
@@ -170,6 +170,24 @@
                 <artifactId>maven-antrun-plugin</artifactId>
                 <executions>
                     <execution>
+                        <id>add-in-place-tool</id>
+                        <phase>prepare-package</phase>
+                        <goals>
+                            <goal>run</goal>
+                        </goals>
+                        <configuration>
+                            <target>
+                                <!-- download file -->
+                                <get src="https://product-dist.wso2.com/downloads/in-place/release/update.zip"
+                                     dest="${project.build.directory}/update.zip"
+                                     verbose="false"
+                                     usetimestamp="true"/>
+                                <unzip src="${project.build.directory}/update.zip"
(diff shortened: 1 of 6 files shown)


distribution/pom.xml
```diff
@@ -169,6 +169,24 @@
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-antrun-plugin</artifactId>
                 <executions>
+                    <execution>
```
praminda: Instead of a antrun execution, can you use maven download plugin with `unpack=true` option to do this.
We are already doing similar execution with download plugin in https://github.com/nihlaakram/product-microgateway/blob/b32387258e236161de282934352d961e0fd92464/distribution/pom.xml#L148

distribution/src/main/assembly/bin.xml
```diff
@@ -82,6 +82,13 @@
                 <exclude>*/**</exclude>
             </excludes>
         </fileSet>
+
+        <!-- Copying the in-place update tools-->
+        <fileSet>
```
praminda: We don't update toolkit via wum. For toolkit updates we do a release and host the dist in website. So I don't think we need to add update tool here

praminda:
@nihlaakram Added few comments to the PR. Can you please send a new PR with the suggested changes. Also please fill the PR templates. 😃

distribution/pom.xml
nihlaakram: This is the convention followed for all products.

distribution/src/main/assembly/bin.xml
nihlaakram: Will send another PR with this.

PR #848, opened 2019-10-18
Fix #847 - Support open API without extensions

### LINKED ISSUE #847: Build Microgateway project without open API vendor extensions
### Describe your problem(s)
The current microgateway version(3.0.2) requires at least two open API vendor extensions to generate the microgateway executable artifact. Those extensions are 
1. x-wso2-basePath  -- specify the base path which is the sub context which categorize all the http operation defined in the open API
2. x-wso2-production-endpoints -- specify the actual implementation url of the endpoint.

We need to support the open APIs predefined attributes.
Based on the open API version these attributes may differ.

### Describe your solution
Open API version 2 : The actual service endpoint is defined using a combination of three attributes[1]. 1. Scheme(http,https) 2. host 3. basePath
Open API version 3 - The backend complete url is defined as array of urls under the attribute servers[2].

[1] - https://swagger.io/docs/specification/2-0/api-host-and-base-path/
[2] - https://swagger.io/docs/specification/api-host-and-base-path/

### How will you implement it
So based on the open API version, if the developer provided open API contains either combination of scheme, host and base path(open API 2) or the server urls(open API 3), micro gateway will not look for the wso2 specific vendor extensions.

---

Rajith90:
### Purpose
Standard open API without wso2 specific vendor extensions should be allow to create micro gateway projects.
From the PR we supports 
Open API v3 - servers object to get the production endpoints and base path
Open API v2 - scheme, host, base path objects to get the base path and production endpoints

### Issues

Fixes #847

### Automation tests
 - Unit tests added: Yes/No

### Tested environments

Not Tested

---
- [x] Assigned 'Type' label
- [x] Assigned the project

### DIFF
--- a/components/micro-gateway-cli/src/main/java/org/wso2/apimgt/gateway/cli/model/template/service/BallerinaPath.java
+++ b/components/micro-gateway-cli/src/main/java/org/wso2/apimgt/gateway/cli/model/template/service/BallerinaPath.java
@@ -16,12 +16,15 @@
 
 package org.wso2.apimgt.gateway.cli.model.template.service;
 
+import io.swagger.v3.oas.models.Operation;
 import io.swagger.v3.oas.models.PathItem;
+import io.swagger.v3.oas.models.servers.Server;
 import org.wso2.apimgt.gateway.cli.exception.BallerinaServiceGenException;
 import org.wso2.apimgt.gateway.cli.model.rest.ext.ExtendedAPI;
 
 import java.util.AbstractMap;
 import java.util.LinkedHashSet;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
@@ -45,36 +48,43 @@ public class BallerinaPath implements BallerinaOpenAPIObject<BallerinaPath, Path
         // Swagger PathItem object doesn't provide a iterable structure for operations
         // Therefore we have to manually check if each http verb exists
         if (item.getGet() != null) {
+            setServersToOperationLevel(item.getGet(), item.getServers());
             operation = new BallerinaOperation().buildContext(item.getGet(), api);
             entry = new AbstractMap.SimpleEntry<>("get", operation);
             operations.add(entry);
         }
         if (item.getPut() != null) {
+            setServersToOperationLevel(item.getPut(), item.getServers());
             operation = new BallerinaOperation().buildContext(item.getPut(), api);
             entry = new AbstractMap.SimpleEntry<>("put", operation);
             operations.add(entry);
         }
         if (item.getPost() != null) {
+            setServersToOperationLevel(item.getPost(), item.getServers());
             operation = new BallerinaOperation().buildContext(item.getPost(), api);
             entry = new AbstractMap.SimpleEntry<>("post", operation);
             operations.add(entry);
         }
         if (item.getDelete() != null) {
+            setServersToOperationLevel(item.getDelete(), item.getServers());
             operation = new BallerinaOperation().buildContext(item.getDelete(), api);
             entry = new AbstractMap.SimpleEntry<>("delete", operation);
             operations.add(entry);
         }
         if (item.getOptions() != null) {
+            setServersToOperationLevel(item.getOptions(), item.getServers());
             operation = new BallerinaOperation().buildContext(item.getOptions(), api);
             entry = new AbstractMap.SimpleEntry<>("options", operation);
             operations.add(entry);
         }
         if (item.getHead() != null) {
+            setServersToOperationLevel(item.getHead(), item.getServers());
             operation = new BallerinaOperation().buildContext(item.getHead(), api);
             entry = new AbstractMap.SimpleEntry<>("head", operation);
             operations.add(entry);
         }
         if (item.getPatch() != null) {
+            setServersToOperationLevel(item.getPatch(), item.getServers());
             operation = new BallerinaOperation().buildContext(item.getPatch(), api);
             entry = new AbstractMap.SimpleEntry<>("patch", operation);
             operations.add(entry);
@@ -95,5 +105,19 @@ public class BallerinaPath implements BallerinaOpenAPIObject<BallerinaPath, Path
 
     public Set<Map.Entry<String, BallerinaOperation>> getOperations() {
         return operations;
+    }
+
+
+    /**
+     * Method to set the Path level server opeartions to operation level servers,
+     * if operation level servers are not present
+     *
+     * @param operation {@link Operation} The current considered operation object
+     * @param pathLevelServers list if server objects defined in the open API PAth level
+     */
+    private void setServersToOperationLevel(Operation operation, List<Server> pathLevelServers) {
+        if (operation.getServers() == null && pathLevelServers != null) {
+            operation.setServers(pathLevelServers);
+        }
     }
 }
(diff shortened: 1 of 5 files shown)


components/micro-gateway-cli/src/main/java/org/wso2/apimgt/gateway/cli/utils/OpenAPICodegenUtils.java
```diff
@@ -354,26 +380,56 @@ private static EndpointListRouteDTO extractEndpointFromOpenAPI(Object endpointEx
...
-        if (openAPIExtensions == null) {
-            throw new CLIRuntimeException(" At least '" + OpenAPIConstants.BASEPATH + "' property and '"
-                    + OpenAPIConstants.PRODUCTION_ENDPOINTS + "' property should present in the open API "
-                    + "definition: '" + openApiFilePath + "'.");
+        String basePath = null;
+        if (openAPIExtensions != null && openAPIExtensions.get(OpenAPIConstants.BASEPATH) != null) {
+            basePath = (String) openAPIExtensions.get(OpenAPIConstants.BASEPATH);
+        } else if (servers != null) {
+            //We  need to check only the first server url to derive the base path. Each and every server url
+            // should contains the same base path.
```
VirajSalaka: There is a confusion here. For each resource and path, we consider the servers object to be the backend. But when it is in the API level, we take that to define the path exposed from the gateway as well. WDYT?

VirajSalaka: IMO, if the user does not provide the basepath via the extension, we should not expose the actual backend's basepath.

components/micro-gateway-cli/src/main/java/org/wso2/apimgt/gateway/cli/utils/OpenAPICodegenUtils.java
```diff
@@ -354,26 +380,56 @@ private static EndpointListRouteDTO extractEndpointFromOpenAPI(Object endpointEx
...
         if (basePath == null || basePath.isEmpty()) {
-            throw new CLIRuntimeException("'" + OpenAPIConstants.BASEPATH + "' property is not included in openAPI " +
-                    "definition '" + openApiFilePath + "'.");
-        }
-        basePath = basePath.startsWith("/") ? basePath : "/" + basePath;
-        if (basePathMap.containsKey(basePath)) {
-            throw new CLIRuntimeException("The value for '" + OpenAPIConstants.BASEPATH + "' " + basePath +
-                    " property is duplicated in the following openAPI definitions.\n" + basePathMap.get(basePath) +
-                    "\n" + openApiFilePath);
+            basePath = "/";
```
VirajSalaka: Having two apis with the same basepath may cause problems in the generated ballerina code. Needs to check that.

components/micro-gateway-cli/src/main/java/org/wso2/apimgt/gateway/cli/model/template/service/BallerinaPath.java
```diff
@@ -96,4 +106,18 @@ public BallerinaPath getDefaultValue() {
...
+
+    /**
+     * Method to set the Path level server opeartions to operation level servers,
+     * if operation level servers are not present
+     *
+     * @param operation {@link Operation} The current considered operation object
+     * @param pathLevelServers list if server objects defined in the open API PAth level
+     */
+    private void setServersToOperationLevel(Operation operation, List<Server> pathLevelServers) {
+        if (operation.getServers() == null && pathLevelServers != null) {
```
VirajSalaka: operation.getExtensions also needs to be considered here

components/micro-gateway-cli/src/main/java/org/wso2/apimgt/gateway/cli/utils/OpenAPICodegenUtils.java
Rajith90: Here, we add the server url base path as the micro gateway base path, and hence get added to every path level operation as well.
We should not enforce x-wso2-basePath , so thats why we need to take the base path from the server url.

components/micro-gateway-cli/src/main/java/org/wso2/apimgt/gateway/cli/model/template/service/BallerinaPath.java
Rajith90: Yes, since we do not support adding path level endpoints, we can add this as an improvement. But in the open API spec servers can be added at path level, so we have to support it. We can add adding extensions also at path level as an improvement. Please create an issue for this.

components/micro-gateway-cli/src/main/java/org/wso2/apimgt/gateway/cli/utils/OpenAPICodegenUtils.java
Rajith90: We maintains the base path map for all the APIS, so befora assigning the base path we check previously added open APIs contains the same base path. This check will fail this during the build time, saying same base path exists in another API also.

components/micro-gateway-cli/src/main/java/org/wso2/apimgt/gateway/cli/model/template/service/BallerinaPath.java
VirajSalaka: Actually my point here is different. It is the logic. If there is an operation with custom extension along with a path level server object, this logic will assign the path level object rather than the value in the custom extension.

components/micro-gateway-cli/src/main/java/org/wso2/apimgt/gateway/cli/utils/OpenAPICodegenUtils.java
VirajSalaka: Let's assume a scenario where two openAPI definitions are provided and both do not have the basepath variable. Then both of them will have the basePath "/". Although the basepath is not mandatory according to this PR, we have to provide the error that the same basepath exists.

components/micro-gateway-cli/src/main/java/org/wso2/apimgt/gateway/cli/utils/OpenAPICodegenUtils.java
VirajSalaka: When there are path level/ operation level endpoints, it will not be an issue. But if the developer is relying on API level endpoints, it is not appropriate to expose the basepath of his backend. IMO, that is a piece of information we should not  expose to the users of API.

components/micro-gateway-cli/src/main/java/org/wso2/apimgt/gateway/cli/model/template/service/BallerinaPath.java
Rajith90: The priority alway given to the extension, If the extension is not present then only the servers object is look at. So if there is an operation level extension ,that value will get chosen, not the path level server value

components/micro-gateway-cli/src/main/java/org/wso2/apimgt/gateway/cli/utils/OpenAPICodegenUtils.java
Rajith90: Yes, there is a warning message is printed on the cases when there is no base path presents and we take default value as "/". The idea is to get build the microgateway without extensions. So the basic scenarios should work without extensions. When the requirements get complicated anyway user will have to use the extensions

components/micro-gateway-cli/src/main/java/org/wso2/apimgt/gateway/cli/utils/OpenAPICodegenUtils.java
Rajith90: I don't think exposing base path as a threat or security concern, It is the backend URL(IP/host and port) that we should not expose to the outside

PR #898, opened 2019-11-29
Fix #896 - Expose token, revoke endpoints in 9095 port

### LINKED ISSUE #896: Expose /token and /revoke endpoints in 9095 port
### Describe your problem(s)
When customers require SSL termination at the microgateway level, it becomes a must to use two domains for token endpoint and normal API endpoints. For example: apis.abc.com and auth.abc.com. It can confuse the users in some use cases where customers need to expose some of their login related endpoints via the gateway (apis.abc.com) which uses the normal API related domain. Then for the login related activities, they have talk cross-domains which is confusing from user PoV.

### Describe your solution
Expose the /token and /revoke endpoints as well in 9095 port with a mechanism to skip the default filters. Then all the endpoints can be exposed in a single domain when SSL termination at the microgateway is required.

Rajith90:
### Purpose
/token, /revoke and etc endpoints are exposed via 9096 port in mgw. Which is different from 9095 port where the api requests are served. We need to expose token services also using 9095 port in scenarios like when the ssl termination happens at the gateway level, then routing to relevant port will be difficult from an load balancer

### Issues

Fixes #896

### Automation tests
 - Unit tests added: No
 - Integration tests added: No

### Tested environments

MacOS

---
- [x] Assigned 'Type' label
- [x] Assigned the project

### DIFF
--- a/components/micro-gateway-cli/src/main/resources/templates/tokenServices.mustache
+++ b/components/micro-gateway-cli/src/main/resources/templates/tokenServices.mustache
@@ -12,7 +12,10 @@ import wso2/gateway;
         }
      {{/corsConfiguration.corsConfigurationEnabled}}
 }
-service authorizeService on tokenListenerEndpoint {
+@gateway:Filters {
+    skipAll: true
+}
+service authorizeService on tokenListenerEndpoint, apiSecureListener {
 
     @http:ResourceConfig {
         path: "/*"
@@ -49,7 +52,10 @@ service authorizeService on tokenListenerEndpoint {
         }
     {{/corsConfiguration.corsConfigurationEnabled}}
 }
-service tokenService on tokenListenerEndpoint {
+@gateway:Filters {
+    skipAll: true
+}
+service tokenService on tokenListenerEndpoint, apiSecureListener {
 
     @http:ResourceConfig {
         path: "/*"
@@ -86,7 +92,10 @@ service tokenService on tokenListenerEndpoint {
         }
     {{/corsConfiguration.corsConfigurationEnabled}}
 }
-service revokeService on tokenListenerEndpoint {
+@gateway:Filters {
+    skipAll: true
+}
+service revokeService on tokenListenerEndpoint, apiSecureListener {
 
     @http:ResourceConfig {
         path: "/*"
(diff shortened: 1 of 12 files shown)


components/micro-gateway-cli/src/main/resources/templates/tokenServices.mustache
```diff
@@ -145,4 +157,11 @@ service userInfoService on tokenListenerEndpoint {
             _ = caller->respond(errorResponse);
         }
     }
+}
+
+public function addTokenServicesFilterAnnotation() {
+    gateway:filterConfigAnnotationMap["authorizeService"] = gateway:getFilterDetailsFromServiceAnnotation(reflect:getServiceAnnotations(authorizeService));
```
praminda: Shall we rename method name to, 
```diff
-- getFilterDetailsFromServiceAnnotation
++ getFilterConfigFromServiceAnnotation
```

components/micro-gateway-core/src/main/ballerina/gateway/annotation.bal
```diff
@@ -31,4 +31,10 @@ public type APIConfiguration record {
 
 public annotation <service> API APIConfiguration;
 
+public type FilterConfiguration record {
+    boolean skipAll = false ;
```
praminda:
```suggestion
    boolean skipAll = false;
```

components/micro-gateway-cli/src/main/resources/templates/tokenServices.mustache
```diff
@@ -145,4 +157,11 @@ service userInfoService on tokenListenerEndpoint {
...
+    gateway:filterConfigAnnotationMap["tokenService"] = gateway:getFilterDetailsFromServiceAnnotation(reflect:getServiceAnnotations(tokenService));
+    gateway:filterConfigAnnotationMap["revokeService"] = gateway:getFilterDetailsFromServiceAnnotation(reflect:getServiceAnnotations(revokeService));
+    gateway:filterConfigAnnotationMap["userInfoService"] = gateway:getFilterDetailsFromServiceAnnotation(reflect:getServiceAnnotations(userInfoService));
 }
```
praminda:
```suggestion
}

```

praminda:
Since @Rajith90 is not available I'm going to merge this PR and resolve the comments in a new PR

Issue #880, opened 2019-11-21
Disable CORS configuration by default


Rajith90:
### Description:
In toolkit-config.toml file the cors configurations are enabled by default. This should be disabled by default .

### Affected Product Version:
3.0.2

### Environment details (with versions):
- OS:
- Client:
- Env (Docker/K8s):

---

Issue #860, opened 2019-11-11
Support CORS configuration for /token API


malinthaprasan:
### Describe your problem(s)
Currently, token API in microgateway doesn't include CORS configuration related annotations. A user needs to manually edit the token API bal file and put the CORS annotations. This is a common requirement when someone building SPAs consuming micro gateway APIs. The SPA might need to access the token endpoint for generating tokens and refreshing. If they use a grant type like implicit which is a common grant type for SPAs, there's unlikely they will use a proxy for the token endpoint

### Describe your solution
Making this configuration available OOTB so that the user doesn't need to edit any ballerina file.

### How will you implement it
N/A

PR #909, opened 2019-12-10
Fix #879 #880 jballerina - CORS support for token services and improve reset command

### LINKED ISSUE #879: reset command completely resets the toolkit-config.toml file
### Description:
The idea of the reset command is to prompt a different user login when importing APIs from API Manager. Right now when reset command is executed it resets all the configurations. So the users will miss the previously specified urls, cors configurations and all.

### Steps to reproduce:
successfully import api from API manager.
Then check the toolkit-config.toml
execute the reset command. Then complete file will be replaced with default one

### Affected Product Version:
3.0.2

- OS:
- Client:
- Env (Docker/K8s):

---

Rajith90:
### Purpose
Following three fixes are addressed in this PR 
1. Currently the microgateway does not allows a way to specify the CORS configurations for /token, /authorize, and other proxy endpoints. With this PR it can be configured in the toolkit-config.toml file of the micr gateway toolkit.

`
[corsConfiguration] corsConfigurationEnabled = true accessControlAllowCredentials = false accessControlAllowOrigins = ["example.com"] accessControlAllowHeaders = ["authorization", "Access-Control-Allow-Origin", "Content-Type"] accessControlAllowMethods = ["GET", "PUT", "POST", "DELETE", "PATCH"]
`

2. Reset command completely resets the toolkit-config.toml. But it should only resets the user credentials only.

3. Disable cors configuration enabled in toolkit-config.toml . By default the cors config will be disabled

### Issues

Fixes #879 #880 #860

### Automation tests
 - Unit tests added: Yes/No

### Tested environments

Not Tested

---
- [x] Assigned 'Type' label
- [x] Assigned the project


components/micro-gateway-cli/src/main/resources/templates/tokenServices.mustache
```diff
@@ -0,0 +1,153 @@
...
+            json errMsg = { "error": "error occurred while invoking the user info endpoint" };
+            errorResponse.setJsonPayload(errMsg);
+            forwardedResponse = errorResponse;
+        }
+        var result = caller->respond(forwardedResponse);
+        if (result is error) {
+           log:printError("Error when responding during the user info endpoint request", err = result);
+        }
+    }
+}
```
praminda:
```suggestion
}

```

PR #955, opened 2020-01-24
Fix #954 - Fix cache initialisation issue

### LINKED ISSUE #954: Token caches initialisation binds to default values
### Description:
Due to defining the cache objects globally, and using those cache objects within handlers which are also used when defining global listeners. So due to this nature token cache objects get initialised with default values when defining handlers in order initialise the listeners

### Affected Product Version:

3.1.0-SNAPSHOT

- OS:
- Client:
- Env (Docker/K8s):

---

Rajith90:
### Purpose
Fixes the issue of cache configs initialisation with default values even we provide values with the configuration.

### Issues

Fixes #954 

### Automation tests
 - Unit tests added: Yes/No

### Tested environments

Not Tested

---
- [x] Assigned 'Type' label
- [x] Assigned the project

### DIFF
--- a/components/micro-gateway-core/src/main/ballerina/src/gateway/listeners/api_gateway_listener.bal
+++ b/components/micro-gateway-core/src/main/ballerina/src/gateway/listeners/api_gateway_listener.bal
@@ -33,11 +33,16 @@ public type APIGatewayListener object {
     public http:Listener httpListener;
 
     public function __init(int port, http:ListenerConfiguration config) {
+        // Since http listeners is wrapped inside https listener also, this init method get invoked twice per
+        // each listener. This check will make sure that configurations are read only once and respective
+        //objects are initialized only once.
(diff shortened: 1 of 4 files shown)


components/micro-gateway-core/src/main/ballerina/src/gateway/listeners/api_gateway_listener.bal
```diff
@@ -68,16 +71,21 @@ public type APIGatewayListener object {
     public function __detach(service s) returns error? {
         return self.httpListener.__detach(s);
     }
-
-
 };
 
+function initiateAuthenticationHandlers(http:ListenerConfiguration config) {
+    http:ListenerAuth auth = {
```
praminda: Let's add a comment on why we moved this logic from `listener.mustache` to here

PR #1047, opened 2020-02-27
Change ballerina base dist and docker image

praminda:
### Purpose

- Remove unused artifacts
OS independent runtime distribution is no longer needed. Therefore removed it from the source.
- Change wum tool download
wum tool is always downloaded when ant plugin is used to download wum tool. This is now done with download plugin to make use of m2 cache
- Download b7a from dist
maven ballerina distribution artifacts doesn't contain b7a docker, k8s libs. Therefore we need to download the b7a dist from website.
- Update base image 
Used a lightweight [alpine based image](https://github.com/Docker-Hub-frolvlad/docker-alpine-glibc) to reduce the image size.
### Issues

Task

### Automation tests
 - Unit tests added: No
 - Integration tests added: No

### Tested environments

MacOS

---
- [x] Assigned 'Type' label
- [x] Assigned the project


docker/Dockerfile
```diff
@@ -41,47 +41,78 @@ ARG MOTD='printf "\n\
...
+    echo \
+        "-----BEGIN PUBLIC KEY-----\
+        MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEApZ2u1KJKUu/fW4A25y9m\
+        y70AGEa/J3Wi5ibNVGNn1gT1r0VfgeWd0pUybS4UmcHdiNzxJPgoWQhV2SSW1JYu\
+        tOqKZF5QSN6X937PTUpNBjUvLtTQ1ve1fp39uf/lEXPpFpOPL88LKnDBgbh7wkCp\
+        m2KzLVGChf83MS0ShL6G9EQIAUxLm99VpgRjwqTQ/KfzGtpke1wqws4au0Ab4qPY\
+        KXvMLSPLUp7cfulWvhmZSegr5AdhNw5KNizPqCJT8ZrGvgHypXyiFvvAH5YRtSsc\
+        Zvo9GI2e2MaZyo9/lvb+LbLEJZKEQckqRj4P26gmASrZEPStwc+yqy1ShHLA0j6m\
+        1QIDAQAB\
+        -----END PUBLIC KEY-----" | sed 's/   */\n/g' > "/etc/apk/keys/sgerrand.rsa.pub" && \
```
Rajith90: Is this key has a expiry time

Rajith90: Is this key has a expiry time

praminda: This is the public key for the glibc for alpine software repository. This might expire but that doesn't effect the released docker images. This is just a build time dependency for fetching the required glibc packages. After that there is no use of this cert. So if it expires we can update it to latest when we are building the latest image.

PR #1076, opened 2020-02-29
Fix issues in combined auth schemes

### LINKED ISSUE #1074: Application security optional and adding default auth schemes
### Description:

When application security is optional, do not add default auth schemes.
When application security optional, mutual ssl authentication should be in OR combination with other application securities.
If application security optional is not defined at the resource level, take API level configuration for application security optional. 
### Affected Product Version:

3.1.0 beta

- OS:
- Client:
- Env (Docker/K8s):

---

AmaliMatharaarachchi:
### Purpose

Add more debug logs
Fix issues in combined auth schemes

### Issues

Fixes #1074

### Automation tests
 - Unit tests added: Yes/No

### Tested environments

Not Tested

---
- [x] Assigned 'Type' label
- [x] Assigned the project

### DIFF
--- a/components/micro-gateway-core/src/main/ballerina/src/gateway/utils/utils.bal
+++ b/components/micro-gateway-core/src/main/ballerina/src/gateway/utils/utils.bal
@@ -944,6 +944,9 @@ function appendMultipleJWTIssuers(http:InboundAuthHandler[] handlers) {
 # + return - auth handlers
 public function getHandlers(string[] appSecurity) returns http:InboundAuthHandler[] {
     http:InboundAuthHandler[] handlers = [];
+    if (isDebugEnabled) {
+        printDebug(KEY_UTILS, "Adding auth handlers : " + appSecurity.toString());
+    }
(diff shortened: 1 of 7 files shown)


components/micro-gateway-core/src/main/ballerina/src/gateway/utils/utils.bal
```diff
@@ -931,6 +931,9 @@ function appendMultipleJWTIssuers(http:InboundAuthHandler[] handlers) {
 # + return - auth handlers
 public function getHandlers(string[] appSecurity) returns http:InboundAuthHandler[] {
     http:InboundAuthHandler[] handlers = [];
+    if (isDebugEnabled) {
```
Rajith90: No need of the check

AmaliMatharaarachchi: I added it since there is a string concat operation and an array to string conversion.

PR #1114, opened 2020-03-08
Improve error logs on extensions

### LINKED ISSUE #1111: Improve error logs on api definition extensions
### Describe your problem(s)

When I set x-wso2-application-security as optional on a resource. During the project build, it throws below error which is acceptable as I don’t have mutual ssl configured. 

`micro-gw: Application security is given as optional but Mutual SSL is not mandatory for the API`

But it's better to give which operation which made the application security optional.
### How will you implement it

---

AmaliMatharaarachchi:
### Issues

Fixes #1111

### Automation tests
 - Unit tests added: No
 - Integration tests added: No

### Tested environments

Tested

---
- [x] Assigned 'Type' label
- [x] Assigned the project

### DIFF
--- a/components/micro-gateway-cli/src/main/java/org/wso2/apimgt/gateway/cli/utils/OpenAPICodegenUtils.java
+++ b/components/micro-gateway-cli/src/main/java/org/wso2/apimgt/gateway/cli/utils/OpenAPICodegenUtils.java
@@ -831,8 +831,8 @@ public class OpenAPICodegenUtils {
             logger.debug("Getting Application security by the extension for API '" + openAPI.getInfo().getTitle()
                     + "' version '" + openAPI.getInfo().getVersion() + "'");
         }
-        ApplicationSecurity appSecurityfromDef = populateApplicationSecurity(openAPI.getInfo().getTitle(),
-                openAPI.getInfo().getVersion(), openAPI.getExtensions(), api.getMutualSSL());
+        ApplicationSecurity appSecurityfromDef = populateApplicationSecurity(api.getName(), api.getVersion(),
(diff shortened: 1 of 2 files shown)


components/micro-gateway-cli/src/main/java/org/wso2/apimgt/gateway/cli/utils/OpenAPICodegenUtils.java
```diff
@@ -827,8 +831,8 @@ private static void setMgwAPISecurityAndScopes(ExtendedAPI api, OpenAPI openAPI)
             logger.debug("Getting Application security by the extension for API '" + openAPI.getInfo().getTitle()
                     + "' version '" + openAPI.getInfo().getVersion() + "'");
         }
-        ApplicationSecurity appSecurityfromDef =
-                populateApplicationSecurity(openAPI.getExtensions(), api.getMutualSSL());
+        ApplicationSecurity appSecurityfromDef = populateApplicationSecurity(openAPI.getInfo().getTitle(),
```
praminda: I'm not that happy about passing openapi title and version as the API name/version to this method. This might make some confusion in the future if we have to add some API name dependent code to this method. (OpenAPI title and actual API name is bit different in MGW - ie. We remove spaces from the title).
Do we have any other options here?

AmaliMatharaarachchi: Fixed in https://github.com/wso2/product-microgateway/pull/1114/commits/ee525e25fc7b837711cd0b4f825506108b968ab1

PR #1115, opened 2020-03-08
Fix #1113 - Template the API base path with version

### LINKED ISSUE #1113: Provide a way to template the base path extension of open API
### Describe your problem(s)
Currently the x-wso2-basePath extension value is directly used as the context to expose the API. 
There should be way to append/prepend or include open API version also into the base path in which the API is exposed

### Describe your solution
Introduce a new extension called **x-wso2-context-template** in order to template the base path of the API

### How will you implement it

---

Rajith90:
### Purpose
New open API extension is introduced to template the base path of the API. This is explained in the issue #1113 

### Issues

Fixes #1113 

### Automation tests
 - Unit tests added: Yes/No

### Tested environments

Not Tested

---
- [x] Assigned 'Type' label
- [x] Assigned the project

### DIFF
--- a/components/micro-gateway-cli/src/main/java/org/wso2/apimgt/gateway/cli/utils/OpenAPICodegenUtils.java
+++ b/components/micro-gateway-cli/src/main/java/org/wso2/apimgt/gateway/cli/utils/OpenAPICodegenUtils.java
@@ -391,7 +391,7 @@ public class OpenAPICodegenUtils {
         api.setEndpointConfigRepresentation(mgwEndpointConfigDTO);
 
         setMgwAPISecurityAndScopes(api, openAPI);
-        api.setSpecificBasepath(extensions.get(OpenAPIConstants.BASEPATH).toString());
+        api.setSpecificBasepath(resolveBasePathFromContextTemplate(extensions, openAPI.getInfo().getVersion()));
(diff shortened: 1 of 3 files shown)


components/micro-gateway-cli/src/main/java/org/wso2/apimgt/gateway/cli/utils/OpenAPICodegenUtils.java
```diff
@@ -1175,4 +1175,22 @@ public static void addDefaultAuthProviders(List<String> authProviders) {
         authProviders.add(OpenAPIConstants.APISecurity.oauth2.name());
         authProviders.add(OpenAPIConstants.APISecurity.jwt.name());
     }
+
+    private static String resolveBasePathFromContextTemplate(Map<String, Object> extensions, String version) {
+        String basePath = extensions.get(OpenAPIConstants.BASEPATH).toString();
```
AmaliMatharaarachchi: Doesn't this cause NPE if basepath extension is not defined?

Rajith90: There is basepath validation method which checks and set  the base path to the extension. So at this point extension is always available

PR #1209, opened 2020-04-22
Fix #1207 - Issue when endpoint url contain special characters

### LINKED ISSUE #1207: Open API production endpoint having special characters cause request failures
### Description:
If the x-wso2-production-endpoint or server object urls contains special characters, it will proxy the request to wrong endpoint
### Steps to reproduce:
1. Define the production endpoint as below in the open API definition
```yaml
x-wso2-endpoints:
- 3b8e54f729cb697e9fecdd36e4e329bb:
    urls:
    - 'https://postman-echo.com/post?foo1=bar1&foo2=bar2'
    type: http
```
2. Then build and invoke the gateway
3. This will forward the request to the endpoint https://postman-echo.com/post?foo1=
### Affected Product Version:
3.1.0

- OS:
- Client:
- Env (Docker/K8s):

---

### LINKED ISSUE #1210: Interceptor get request path parameters and the matching resource path
### Describe your problem(s)
In both java and ballerina interceptors right now there is no way to get the path parameters in the interceptor level. We should provide a way to get path parameters as a map.
For ex if the resource is /pet/{petID} and the request comes with /pet/456-567
then map should provide petId->456-567

And also the details like matching resource, API Name, version like details are not available to the interceptors. 
We should set these details to the invocation context so the users can read them.

### Describe your solution
1. Provide a method for java interceptors to get the path parameters.
2. In ballerina interceptors this can be read from invocation context as a map
3. And request details will be set to invocation context so in both java and ballerina interceptors can read them.

---

Rajith90:
### Purpose
When production endpoints containing special characters are specified in the open API definition, microgateway route the
request to a wrong endpoint. This issue happens for production endpoints like "http://test.com?value1=foo&valu2=bar"
where it contains query parameters.

### Issues

Fixes #1207 
Fixes #1210 

### Automation tests
 - Unit tests added: No

### Tested environments

Not Tested

---
- [x] Assigned 'Type' label
- [x] Assigned the project


tests/src/test/resources/openAPIs/common_api.yaml
```diff
@@ -121,6 +121,49 @@ paths:
...
+          explode: true
+          schema:
+            type: array
+            items:
+              type: string
+              enum:
+                - available
+                - pending
+                - sold
+              default: available
```
praminda, line 148: Just to keep things simple
```suggestion
```

tests/src/test/java/org/wso2/micro/gateway/tests/grpc/GrpcServer.java
```diff
@@ -31,7 +31,7 @@
 
     public void start() throws IOException {
         /* The port on which the server should run */
-        int port = 50051;
+        int port = 50075;
```
praminda, line 34: Any reason to change the port? If there is an specific reason, better if you can put that in the PR description. It might be useful later.

Rajith90: There seems to be port conflict with mac. Its better to use a different port

PR #1228, opened 2020-05-06
Fix multiple jms subscription

### LINKED ISSUE #1227: Second MGW instance fails to subcribe to JMS topic
### Description:

When MGW is configured for distributed throttling only one instance can subscribe to Traffic Manager's JMS topic. Second instance fails to subscribe.
```
[ThrottleUtil] [-] Error while creating the jms subscriber. : error javax.jms.JMSException message=Error registering consumer: org.wso2.andes.AMQChannelClosedException: Error: Cannot subscribe to queue carbon:sub-1 as it already has an existing exclusive consumer [error code 403: access refused] cause=error org.wso2.andes.AMQChannelClosedException message=Error: Cannot subscribe to queue carbon:sub-1 as it already has an existing exclusive consumer
```
### Steps to reproduce:
1. Enable the `enabledGlobalTMEventPublishing` in more than one MGW instances
2. Subscribe both instances to same Traffic Manager instance.
3. Start TM and MGW instances.
4. Observe the MGW logs.

### Affected Product Version:

3.1.0
- OS: Any

praminda:
### Purpose

There was an issue when multiple JMS subcribers (two mgw nodes) trying to connect to same TM for distributed throttling. Changing the subscriber type fixes the issue.
### Issues

Fixes #1227 

### Automation tests
 - Unit tests added: No
 - Integration tests added: No

### Tested environments

MacOS

---
- [x] Assigned 'Type' label
- [x] Assigned the project

### DIFF
--- a/components/micro-gateway-core/src/main/ballerina/src/gateway/services/throttle_event_listener.bal
+++ b/components/micro-gateway-core/src/main/ballerina/src/gateway/services/throttle_event_listener.bal
@@ -84,7 +84,7 @@ public function startSubscriberService() returns @tainted jms:MessageConsumer | 
             return session;
         } else {
             jms:Destination dest = check session->createTopic("throttleData");
-            jms:MessageConsumer | error subscriberEndpoint = trap session->createDurableSubscriber(dest, "sub-1");
+            jms:MessageConsumer | error subscriberEndpoint = trap session->createConsumer(dest, "");
             if (subscriberEndpoint is error) {
                 printError(KEY_THROTTLE_UTIL, "Error while creating the jms subscriber.", subscriberEndpoint);
             } else {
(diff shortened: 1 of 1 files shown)


components/micro-gateway-core/src/main/ballerina/src/gateway/services/throttle_event_listener.bal
```diff
@@ -84,7 +84,7 @@ public function startSubscriberService() returns @tainted jms:MessageConsumer |
             return session;
         } else {
             jms:Destination dest = check session->createTopic("throttleData");
-            jms:MessageConsumer | error subscriberEndpoint = trap session->createDurableSubscriber(dest, "sub-1");
+            jms:MessageConsumer | error subscriberEndpoint = trap session->createConsumer(dest, "");
```
Rajith90, line 87: Why we need to pass empty string. I think we can use
```java
createConsumer(dest);
```

praminda: Fixed

PR #1245, opened 2020-05-21
Add log4j logger for native implementation

### LINKED ISSUE #1244: Provide the error logs with its stacktrace
### Describe your problem(s)
When an exception thrown from the native implementation, it does not provide the related stacktrace. (only the error message will be printed)

### Describe your solution
Add log4j2 logger for the logging purposes of the native implementation. 

### How will you implement it
Add log4j2 logger for the logging purposes of the native implementation. 

###Remarks
- For the callhome class, this logging implementation is not added. It will use the ballerina logging mechanism.

---

VirajSalaka:
### Purpose
Provide detailed error logs via using the log4j2 logger for the native implementation.

### Issues

Fixes #1244 

### Automation tests
 - Unit tests added: No
 - Integration tests added: No

### Tested environments

MacOS Mojave

### Remarks
- To debug the native implementation, the user needs to edit the existing log4j.properties file in <Microgateway-Runtime>/conf/log4j2.properties. (providing --b7a.log.level parameter does not affect on this)
- If the log4j2.properties file does not exist in the <Microgateway-Runtime>/conf, the default configuration would be used.

---
- [x] Assigned 'Type' label
- [x] Assigned the project


distribution/resources/bin/gateway
```diff
@@ -283,6 +283,13 @@ else
     done
 fi
 
+# setup log4j configuration file to debug native implementation
+LOG4J_CONFIGURATION_FILE="${GW_HOME}/conf/log4j2.properties"
+if [ -f "$LOG4J_CONFIGURATION_FILE" ]
+then
```
praminda, line 289:
```suggestion
```

distribution/resources/bin/gateway
```diff
@@ -283,6 +283,13 @@ else
     done
 fi
 
+# setup log4j configuration file to debug native implementation
+LOG4J_CONFIGURATION_FILE="${GW_HOME}/conf/log4j2.properties"
+if [ -f "$LOG4J_CONFIGURATION_FILE" ]
```
praminda, line 288:
```suggestion
if [ -f "$LOG4J_CONFIGURATION_FILE" ]; then
```

distribution/resources/conf/log4j2.properties
```diff
@@ -0,0 +1,15 @@
...
+# settings for the console appender
+appender.console.type = Console
+appender.console.name = STDOUT
+appender.console.layout.type = PatternLayout
+appender.console.layout.pattern = %-4r %d [%t] %-5p [%c] %x - %m%n
+
+# To debug the native implementation
+#loggers = classLevelLogger
+#logger.classLevelLogger.name = org.wso2.micro.gateway.core.{ClassName}
+#logger.classLevelLogger.level = debug
```
praminda, line 15:
```suggestion
#logger.classLevelLogger.level = debug

```

VirajSalaka:
change the service.mustache logs
check if we can print the method name

VirajSalaka:
> change the service.mustache logs
> check if we can print the method name

Method name cannot be logged as it may cause a performance hit.
Mustache files are changed to follow the same log pattern via https://github.com/wso2/product-microgateway/pull/1245/commits/a765c565c3c7b9728f17efe5133657a7e2d9a355

PR #1247, opened 2020-05-22
Support event publishing to multiple Traffic Managers

### LINKED ISSUE #1246: Publishing events to multiple Traffic Manager instances.
### Describe your problem(s)
At the moment, the microgateway only allows publishing events to a single traffic manager endpoint. 

### Describe your solution
Creating a TCP connection for multiple traffic managers and publish events via that connection. 

### How will you implement it
Used the existing implementation from the binary agent source code in [wso2 analytics-common ](https://github.com/wso2/carbon-analytics-common)repository after removing unnecessary dependencies. This native implementation will be called from the ballerina gateway.

### Remarks
The previous default behavior should remain the same. (Event publishing will happen via HTTP).

---

VirajSalaka:
### Purpose
In the global throttling scenario, users may need to publish events to multiple traffic managers.
From this implementation, the users can configure the microgateway to have multiple traffic manager endpoints. Inside the microgateway, it will create a binary connection to those traffic managers and publish the throttle events.  

### Issues

Fixes #1246 

### Automation tests
 - Unit tests added: No
 - Integration tests added: No

### Tested environments

MacOS Mojave

### Remarks
Default behavior would not change. If the user needs to switch from the default HTTP event publishing to binary publisher, the configuration should be provided.

```
[throttlingConfig]
  enabledGlobalTMEventPublishing = true
  [throttlingConfig.binary]
    enabled = true
```
---
- [x] Assigned 'Type' label
- [x] Assigned the project


VirajSalaka:
Please note that the following PR needs to be merged in order to function this.
https://github.com/wso2/product-microgateway/pull/1245

components/micro-gateway-cli/src/main/resources/templates/main.mustache
```diff
@@ -25,6 +25,7 @@ public function main() {
     addTokenServicesFilterAnnotation();
     initThrottlePolicies();
     gateway:initThrottleDataPublisher();
```
VirajSalaka, line 27: if global throttling is initialized no need to have this pool

distribution/resources/conf/default-micro-gw.conf.template
```diff
@@ -246,6 +246,56 @@
     password = "admin"
 
 
+  # Throttling configurations related to event publishing using a binary connection
+  [throttlingConfig.binary]
+    enabled = false
```
VirajSalaka, line 251: set enable true by default

components/micro-gateway-cli/src/main/resources/templates/main.mustache
VirajSalaka: already implemented.

distribution/resources/conf/default-micro-gw.conf.template
VirajSalaka: fixed

PR #1304, opened 2020-06-29
Implement Log4j Implementation as a custom ConfigurationFactory

### LINKED ISSUE #1281: Default Log4j2 configurations are not applied.
### Description:
The log4j configurations which are set programmatically are not applied. Probably due to been overwritten by some other class. 

### Steps to reproduce:
Add a java interceptor to the swagger file and do not include the relevant jar file. This would result in a CNF exception. It can be confirmed via checking the log pattern.  

### Affected Product Version:
Current Master Branch

- OS: MacOS Mojave
- Client:
- Env (Docker/K8s):

---

VirajSalaka:
### Purpose
Rather than using the configurator, here we implement the log4j implementation as a custom configuration factory. During the startup (in gateway script), we provide the implemented class within the microgateway-core as a JVM system variable. 
`-Dlog4j.configurationFactory=org.wso2.micro.gateway.core.logging.MgwLog4j2ConfigurationFactory`

### Issues
Fixes #1281 

### Automation tests
 - Unit tests added: No
 - Integration tests added: No

### Tested environments

MacOS Mojave

---
- [x] Assigned 'Type' label
- [x] Assigned the project


VirajSalaka:
The issue with the configurator approach (the existing one) is that the configurations are overridden by the default log4j2 configuration (after the PR is merged). 
And I have tried using the plugin annotation, instead of adding the configuration class to the script. But it turns out that within the microgateway-core jar file, log4j2plugins.dat contains as it is assigned in the pom. But when the shaded jar is generated, those data are not available in its log4j2plugins.dat. 
Hence, I came up with this solution.

PR #1334, opened 2020-07-16
Fix #1333 - Improvements for subscription validation logic

### LINKED ISSUE #1333: Subscription validation failures due to incorrect subscription state when validate with data stores
### Description:
When connecting with APIM event hub to retrieve subscription data , during the subscription event creation flow , microgateway received two messages for sub creation event. One is with subscription state as "ON_HOLD" and other is as "UN_BLOCKED". If these two events are received in reverse order due to topic subscription messages are not guaranteed to be delivered in same order, then token taken with valid subscription will fail. This is due to subscription state is stored as "ON_HOLD" in data stores. 
Then gateway will respond with error code `900909` with message `The subscription to the API is inactive`
This error is only relevant to 3.2.0-alpha and snapshots only as previous version of mgw (including 3.1.0) does not have the concept of data stores.

### Steps to reproduce:
This error happens intermittently when we subscribe new app to a API and invoke using a token.

### Affected Product Version:
3.2.0-alpha

- OS:
- Client:
- Env (Docker/K8s):

---

Rajith90:
### Purpose
Fix subscription validation issues
Update toolkit REST and DCR versions to work with APIM 3.2.0 by default

### Issues

Fixes #1333 

### Automation tests
 - Unit tests added: No

### Tested environments

Not Tested

---
- [x] Assigned 'Type' label
- [x] Assigned the project

### DIFF
--- a/distribution/resources/cli-conf/toolkit-config.toml
+++ b/distribution/resources/cli-conf/toolkit-config.toml
@@ -3,8 +3,8 @@ httpRequestTimeout = 1000000
 
 [token]
 baseURL = ""
-restVersion = "v1.1"
-dcrVersion = "v0.16"
+restVersion = "v1.2"
+dcrVersion = "v0.17"
 publisherEndpoint = "{baseURL}/api/am/publisher/{restVersion}"
 adminEndpoint = "{baseURL}/api/am/admin/{restVersion}"
 registrationEndpoint = "{baseURL}/client-registration/{dcrVersion}/register"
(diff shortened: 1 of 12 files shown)


distribution/resources/cli-conf/toolkit-config.toml
```diff
@@ -3,8 +3,8 @@ httpRequestTimeout = 1000000
 
 [token]
 baseURL = ""
-restVersion = "v1.1"
-dcrVersion = "v0.16"
+restVersion = "v1.2"
```
AmaliMatharaarachchi, line 6: Shall we updated the default rest and dcr version in here as well. https://github.com/wso2/product-microgateway/blob/9603fbd554d200cca33c10114f70232dbe9b7b6c/components/micro-gateway-cli/src/main/java/org/wso2/apimgt/gateway/cli/constants/RESTServiceConstants.java#L47

Rajith90: fixed with : ba4c1481c18b7e6da52a09bc16141f69f674c3cc

PR #1367, opened 2020-08-04
Fix JWT revocation not working with APIM 3.1.0

### LINKED ISSUE #1364: JWT Revocation does not work with APIM 3.1.0
### Description:
Revoked JWT's are considered valid in Microgateway

### Steps to reproduce:
- Configure JWT revocation
- Create, Build and Publish an API using Microgteway
- Generate a JWT and invoke the API
- Revoke the token.
- Invoke the API again with the same token.

API Invocation successful.

- OS:
- Client:
- Env (Docker/K8s):

---

### LINKED ISSUE #1369: JWT Generation fails due to an error in keystore path in Windows
### Description:
When generating a JWT token in windows, it takes the keystore path without `"\"` and due to that JWT generation fails.
Screenshot (198)

### Steps to reproduce:
Follow the steps in this doc for generating a jwt token with custom claims: https://github.com/wso2/docs-mg/blob/27c6915d63818cb3c6dd6bee0669fa821559374b/en/docs/how-tos/generating-backend-jwt.md

### Affected Product Version:
3.2.0 beta

- OS: Windows 10 Pro
- Client:
- Env (Docker/K8s):

---

menakaj:
### Purpose
This PR fixes multiple issues.

1. JWT revocation does not work with APIM 3.1.0 due to the below reasons.

- Payload parameters are different
- JMS payloads are different
This PR fixes this issue and enables to use JWT revocation with APIM 3.1.0 and MGW 3.2.0

2. JWT generation fails due to path issues in Windows.

### Issues
Fixes #1364
Fixes #1369

### Automation tests
Unit tests added: Yes/No
Integration tests added: Yes/No

### Tested environments
Not Tested

---
- [x] Assigned 'Type' label
- [x] Assigned the project

### DIFF
--- a/components/micro-gateway-cli/src/main/java/org/wso2/apimgt/gateway/cli/cmd/ImportCmd.java
+++ b/components/micro-gateway-cli/src/main/java/org/wso2/apimgt/gateway/cli/cmd/ImportCmd.java
@@ -280,7 +280,7 @@ public class ImportCmd implements LauncherCmd {
             CmdUtils.saveSwaggerDefinitionForMultipleAPIs(projectName, apis,
                     !restVersion.startsWith(CliConstants.REST_API_V1_PREFIX));
         } catch (Exception e) {
-            throw new CLIInternalException("Exception occurred during codeGeneration process: " + e.getMessage());
+            throw new CLIInternalException("Exception occurred during codeGeneration process", e);
         }
         //todo: check if the files has been changed using hash utils
 
--- a/components/micro-gateway-cli/src/main/java/org/wso2/apimgt/gateway/cli/cmd/ImportCmd.java
+++ b/components/micro-gateway-cli/src/main/java/org/wso2/apimgt/gateway/cli/cmd/ImportCmd.java
@@ -280,7 +280,7 @@ public class ImportCmd implements LauncherCmd {
             CmdUtils.saveSwaggerDefinitionForMultipleAPIs(projectName, apis,
                     !restVersion.startsWith(CliConstants.REST_API_V1_PREFIX));
         } catch (Exception e) {
-            throw new CLIInternalException("Exception occurred during codeGeneration process");
+            throw new CLIInternalException("Exception occurred during codeGeneration process: " + e.getMessage());
         }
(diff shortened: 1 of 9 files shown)


components/micro-gateway-cli/src/main/java/org/wso2/apimgt/gateway/cli/cmd/ImportCmd.java
```diff
@@ -280,7 +280,7 @@ public void execute() {
             CmdUtils.saveSwaggerDefinitionForMultipleAPIs(projectName, apis,
                     !restVersion.startsWith(CliConstants.REST_API_V1_PREFIX));
         } catch (Exception e) {
-            throw new CLIInternalException("Exception occurred during codeGeneration process");
+            throw new CLIInternalException("Exception occurred during codeGeneration process: " + e.getMessage());
```
praminda, line 283: How about passing `e` to get the whole trace?
```suggestion
            throw new CLIInternalException("Exception occurred during codeGeneration process", e);
```

distribution/resources/conf/default-micro-gw.conf.template
```diff
@@ -52,6 +52,8 @@
   serverUrl = "https://localhost:9443"
   # The token endpoint context of the Key Manager server
   tokenContext = "oauth2"
+  # When Microgateway is used with older APIM versions for subscription validation by using KeyValidation service.
+  isLegacyKM = false
```
praminda, line 56: Since we go with `enable` prefix for other configs, shall we rename this config to something like `enableLegacyMode`?

components/micro-gateway-cli/src/main/java/org/wso2/apimgt/gateway/cli/cmd/ImportCmd.java
menakaj: The message is not displaying if we pass the exception

distribution/resources/conf/default-micro-gw.conf.template
menakaj: Will fix this in the new branch

PR #1426, opened 2020-10-07
Basic Deployment with docker-compose

VirajSalaka:
### Purpose
The purpose here is to have an all connected setup to continue development. 

### Remarks
by running, build.sh the developers can have the three containers up and running. (This is for development purposes). If the java source code is changed, a new jar will be created from the maven command and will be mounted to the container running.

The go control plane is built and executed within the golang installed image, hence the size of the image is large. 

If the developer needs to run just the go control plane, envoy.yaml file's IP address for the xds_cluster should be changed to his own IP. Then the pilot segment needs to be removed from the docker-compose file

If the user uses `docker-compose-without-dependencies.yaml` for docker-compose up command, the java filter chain is built inside a maven docker image and executed using maven:exec plugin.

### Issues

Fixes #

### Automation tests
 - Unit tests added: No
 - Integration tests added: No

### Tested environments

Not Tested

---
- [ ] Assigned 'Type' label
- [ ] Assigned the project


envoy.yaml
```diff
@@ -43,3 +30,16 @@ static_resources:
       http2_protocol_options: {}
       name: xds_cluster
```
praminda, line 31: Can you switch the cluster name prop to line 25. it makes it easier to identify the cluster

PR #1473, opened 2020-11-09
Load API from checkrequest info

### LINKED ISSUE #1472: Read API details from Check Request
### Describe your problem(s)

In current implementation both CP and filter chain read the api definition from files to load API details. However this need to be changed and only CP should load APIs details from the source it receives API details. Filter chain should only look at the request details coming from the envoy side to identify exact API and resource level details.

### Describe your solution

Read API details from envoy `checkrequest`'s ContextAttributes map and avoid loading APIs from disk volumes.

### How will you implement it

1. Add API attributes to envoy `checkrequest`'s ContextAttributes map
2. Read `ContextAttributes` from filter chain side and load API details

praminda:
### Purpose

- Filter chain doesn't need to load openapi defs separately. All required details will be attached to the checkrequest from envoy in `ContextAttributes` map
- Use common config file for both CP and filter-chain
### Issues

Fixes #1472

### Automation tests
 - Unit tests added: No
 - Integration tests added: No

### Tested environments

MacOS
Docker 19.03.12

---
- [x] Assigned 'Type' label
- [x] Assigned the project


filter-core/src/main/java/org/wso2/micro/gateway/filter/core/api/RestAPI.java
```diff
@@ -85,63 +74,22 @@ public APIConfig getAPIConfig() {
...
-                logger.error("Open API server url templating is used for the url : " + url
-                        + ". But default values is not provided for the variable '" + matcher.group(0)
-                        + "'. Hence correct url will not be resolved during the runtime "
-                        + "unless url is overridden during the runtime");
-            }
-        }
-        return url;
+    private List<ResourceConfig> extractResourceConfig(Map<String, String> attributes) {
+        String resPath = attributes.get(APIConstants.GW_RES_PATH_PARAM);
+        String resMethod = attributes.get(APIConstants.GW_RES_METHOD_PARAM);
```
VirajSalaka, line 85: If multiple http methods are involved with a single path, the filter-core gets a space separated set of Methods (in a pending PR).
Hence shall we split this using space and add a resource for each method.

distribution/src/main/assembly/assembly.xml
```diff
@@ -28,7 +28,7 @@
         </fileSet>
         <fileSet>
             <directory>../resources/conf</directory>
-            <outputDirectory>WSO2-Microgateway-${project.version}/resources/pilot/conf</outputDirectory>
+            <outputDirectory>WSO2-Microgateway-${project.version}/resources/conf</outputDirectory>
```
Rajith90, line 31: Lets use the existing distribution name we used for microgateway runtime

praminda: Fixed with 2921e00

filter-core/src/main/java/org/wso2/micro/gateway/filter/core/api/RestAPI.java
praminda: Fixed with c2b0467

VirajSalaka (approved):

PR #1476, opened 2020-11-10
Resolve Bugs related to routing 

### LINKED ISSUE #1475: Resolve Bugs related to routing based on various aspects
### Describe your problem(s)
The current envoy-microgateway implementation has the following issues.

- When the HTTP or https address is provided without a port value, it looks up for a value called apiDefaultPort. It should be 80 and 443 respectively.
- When multiple methods are provided under a pathItem, the gateway only picks a one value from it.
- Support resource level endpoints.

### Describe your solution

- When there is no port available the port should be hardcoded to 80 (in HTTP scenario) or 443 (in https Scenario)
- The methods need to be provided under header match property of the routes contig as a regular expression. (Since multiple values cannot be provided for that)
-  Servers object and the x-wso2-production-endpoints extension should be used to provide resource level endpoints.

### How will you implement it
Mentioned in the above segment.

---

VirajSalaka:
### Purpose
Address the issues stated in https://github.com/wso2/product-microgateway/issues/1475
Navigate to the remarks sections to critical changes

### Issues

Fixes #1475 

### Automation tests
 - Unit tests added: Yes
 - Integration tests added: No

### Tested environments

Not Tested

### Remarks

- The filter-core will be getting a space-separated HTTP method lists as metadata. Hence the filter-core implementation should have to identify the request's method from the `method` header rather than the metadata sent by envoy.

- The operation level endpoints cannot be supported at the moment, as the envoy routing logic has the minimum granularity level of paths. 

- If `http` and `https` both mentioned under schemes (swagger 2.0) , https will be assigned. If no scheme is provided, it would be `http`

---
- [x] Assigned 'Type' label
- [x] Assigned the project


VirajSalaka:
Test CORS , as it is provided on HTTP Methods.

resources/conf/log_config.toml
```diff
@@ -1,6 +1,6 @@
 # The logging configuration file for control plane
 ######### root Level ########
-logfile = "resources/logs/controlPlane.log"
```
Rajith90, line 3: Whats the purpose of this file

VirajSalaka: Print the control plane logs to a file

Rajith90: I was referring to the log_config.toml file, which is the logging config for the control plane, so thats fine I guess

VirajSalaka: yeah. Just like log4j2 configuration for the filter-core

PR #1518, opened 2020-12-07
support production and sandbox endpoints

### LINKED ISSUE #1495: Support Production Endpoints and Sandbox Endpoints
### Describe your problem(s)
Needs to provide both Production and Sandbox endpoints for a given API resource. This allows the user to test the API functionality by using a sandbox endpoint.

### Describe your solution

The environment type is picked based on the access token. Depending on that value, the proxy should route the request to the production or sandbox endpoints.

### How will you implement it

The environment type is selected after processing the access token. Hence, it is required to set an additional header to the response received from the enforcer (external authz filter) depending on the environment type. [1]

Then the headers_to_metadata http filter needs to be applied to map these header values to some metadata property. The option to remove the additional headers are provided OOTB within the filter. [2]

Finally, the cluster endpoints should be matched based on these metadata values. [3] and [4]

[1] https://www.envoyproxy.io/docs/envoy/v1.16.0/api-v3/service/auth/v3/external_auth.proto.html?highlight=checkresponse#service-auth-v3-okhttpresponse

[2] https://www.envoyproxy.io/docs/envoy/v1.16.0/api-v3/extensions/filters/http/header_to_metadata/v3/header_to_metadata.proto

[3] https://www.envoyproxy.io/docs/envoy/v1.16.0/api-v3/config/endpoint/v3/endpoint_components.proto#envoy-v3-api-msg-config-endpoint-v3-lbendpoint

[4] https://www.envoyproxy.io/docs/envoy/v1.16.0/api-v3/config/cluster/v3/cluster.proto#config-cluster-v3-cluster-lbsubsetconfig

---

VirajSalaka:
### Purpose
Support production and sandbox environment endpoints. The details are explained within the issue.

### Issues

Fixes #1495 

### Automation tests
 - Unit tests added: Yes
 - Integration tests added: No

### Tested environments

macOS Mojave

### Remarks
Needs to check the performance impact. (Since the external Auth filter could add some other header like Authorization)
Needs to document the possible performance impact so that users can see prior to their deployment.

---
- [ ] Assigned 'Type' label
- [ ] Assigned the project


VirajSalaka:
added via a different PR.

Issue #1681, opened 2021-03-04
Support Endpoint Prototype APIs 


HiranyaKavishani:
### Describe your problem(s)
Current envoy support cannot deploy Endpoint prototype APIs

### Describe your solution
Changes needs to be added prior to lifeCycleStatus=Prototyped

### How will you implement it
Endpoints needs to be extracted from api.json file as it doesn't perform at swagger and add it to existing MGWSwagger model

---

dinusha92:
Closing the issue since PR #1682 has fixed this.

PR #1682, opened 2021-03-04
Support Prototype APIs 

HiranyaKavishani:
### Purpose
This PR will add the support to deploy prototype APIs(endpointImplementationType=ENDPOINT) at envoy

### Issues

Fixes https://github.com/wso2/product-microgateway/issues/1681

### Automation tests
 - Unit tests added: Yes/No

### Tested environments

Mac, JDK 11

---
- [x] Assigned 'Type' label
- [x] Assigned the project


VirajSalaka:
Check if the API is pulled if the event is received after startup? I think it is missing here.

VirajSalaka:
Test the following flow. 
- Create an API in publisher (Created state)
- See if the API is applied in the router
- try invoking it. (to see if the API is in the enforcer)

IMO, the API should not be there. If so, create a new issue. (As it is not caused by your pull request)

enforcer/src/main/java/org/wso2/micro/gateway/enforcer/api/API.java
```diff
@@ -38,7 +39,14 @@
 
     default boolean executeFilterChain(RequestContext requestContext) {
         boolean proceed;
-        for (Filter filter : getFilters()) {
+        List<Filter> filterList = getFilters();
+        String lifeCycleState = requestContext.getMathedAPI().getAPIConfig().getApiLifeCycleState();
+        if (lifeCycleState.equals(APIConstants.PROTOTYPED_LIFE_CYCLE_STATUS)) {
```
VirajSalaka, line 44: What would happen if the API is there, but lifecycle state is CREATED?

VirajSalaka, line 44: Move this initialization to the InitFilters method as it happens at the API's arrival point. Then we would not need to run this segment over each request.

HiranyaKavishani:
@VirajSalaka, when event is received after startup same flow is being executed.

enforcer/src/main/java/org/wso2/micro/gateway/enforcer/api/API.java
HiranyaKavishani: When lifecycle state of an API is in CREATED state, that API has not been displayed at devportal. hence  it couldn't be exposed outsider to do the invocations.

HiranyaKavishani: fixing with https://github.com/wso2/product-microgateway/pull/1682/commits/e8719a3b77a193a9c3053de0261b89eee8a8061c

VirajSalaka: yeah, You are right.

VirajSalaka:
> @VirajSalaka, when event is received after startup same flow is being executed.

could you please verify the following flow.
1. API is deployed while it is in CREATED state.
2. Change lifecycle state to Prototyped.
3. Check if it works

I

adapter/internal/api/apis_impl.go
```diff
@@ -115,13 +120,22 @@ func extractAPIProject(payload []byte) (apiProject ProjectAPI, err error) {
...
+				loggers.LoggerAPI.Infof(errMsg)
+				err = errors.New(errMsg)
 				return apiProject, err
 			}
 			apiProject.APIJsn = apiJsn
 			apiProject.APIType = apiType
+			apiProject.APILifeCycleStatus = lifeCycleStatus
+			apiProject.ProductionEndpoint = productionEndpoint
+			apiProject.SandboxEndpoint = sandboxEndpoint
+			apiProject.APIJsn = apiJsn
```
SuKSW, line 138:
```suggestion
		
```
Seems this is an additional line

adapter/internal/api/apis_impl.go
```diff
@@ -202,6 +220,34 @@ func updateAPI(name, version string, apiProject ProjectAPI, environments []strin
 	}
 }
 
+func extractAPIInformation(apiJsn []byte) (apiType string, apiLifeCycleStatus string,
+	productionEndpoint string, sandboxEndpoint string, apiEndpointImplementationType string, err error) {
```
SuKSW, line 224:
```suggestion
func extractAPIInformation(apiJsn []byte) (apiType, apiLifeCycleStatus,
	productionEndpoint, sandboxEndpoint, apiEndpointImplementationType string, err error) {
```
We can update the method signature like this, so it is shorter and also let us be cautious to not mix any because it is visible that all of them are strings.

HiranyaKavishani: Fixed by https://github.com/wso2/product-microgateway/pull/1682/commits/459a83228043b61d1de8f7be99d9a054fefb1cbd

adapter/internal/api/apis_impl.go
HiranyaKavishani: fixed by https://github.com/wso2/product-microgateway/pull/1682/commits/459a83228043b61d1de8f7be99d9a054fefb1cbd

HiranyaKavishani:
@VirajSalaka, I tested the flow. For the prototype APIs, we don't need to handle API_LIFECYCLE_CHANGE=PROTOTYPED event separately as it needs to be deployed every-time once after we do a change on the API. So, DEPLOY_API_IN_GATEWAY event is firing and getting executed same flow.

VirajSalaka (approved):

Issue #1756, opened 2021-03-23
Add Integration tests for Subscription Validation


sajinieKavindya:
### Describe your problem(s)

- Add Integration tests for Subscription validation

---

dinusha92:
Closing as the above PR fixed this.

Issue #1755, opened 2021-03-23
Use API Manager as the Control Plane for Integration tests


sajinieKavindya:
### Describe your problem(s)

- Use API Manager as the Control Plane for Integration tests
- Add utility methods for Mgw test cases that use the APIM as the backend server

### How will you implement it

- Use docker with apim docker-compose.yaml file
- Write utility methods to interact with APIM
---

dinusha92:
Closing as the above PR fixed this.

PR #1757, opened 2021-03-23
Add Integration tests for Subscription Validation

sajinieKavindya:
### Purpose

- Use API Manager as the Control Plane for Integration tests
- Add utility methods for Mgw test cases that use the APIM as the backend server
- Add Integration tests for Subscription validation

### Issues

Fixes https://github.com/wso2/product-microgateway/issues/1755
Fixes https://github.com/wso2/product-microgateway/issues/1756

### Automation tests
 - Unit tests added: N/A
 - Integration tests added: N/A

### Tested environments

Not Tested

---
- [x] Assigned 'Type' label
- [x] Assigned the project

### DIFF
--- a/integration/test-integration/pom.xml
+++ b/integration/test-integration/pom.xml
@@ -44,6 +44,14 @@
                     <suiteXmlFiles>
                         <suiteXmlFile>src/test/resources/testng.xml</suiteXmlFile>
                     </suiteXmlFiles>
+                    <systemProperties>
+                        <property>
+                            <name>framework.resource.location</name>
+                            <value>
+                                ${basedir}/src/test/resources/
+                            </value>
+                        </property>
+                    </systemProperties>
                 </configuration>
             </plugin>
             <plugin>
@@ -219,6 +227,18 @@
             <groupId>org.wso2.am.microgw</groupId>
             <artifactId>org.wso2.micro.gateway.enforcer</artifactId>
             <version>4.0.0-m10-SNAPSHOT</version>
+        </dependency>
+        <dependency>
+            <groupId>org.wso2.am</groupId>
+            <artifactId>org.wso2.am.integration.common.test.utils</artifactId>
+            <version>4.0.0-SNAPSHOT</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.awaitility</groupId>
+            <artifactId>awaitility</artifactId>
+            <version>3.1.2</version>
+            <scope>test</scope>
         </dependency>
     </dependencies>
 
--- a/integration/test-integration/pom.xml
+++ b/integration/test-integration/pom.xml
@@ -231,6 +231,7 @@
         <dependency>
             <groupId>org.wso2.am</groupId>
             <artifactId>org.wso2.am.integration.common.test.utils</artifactId>
+            <!-- TODO:Dinusha - bump the version to 4.0.0-alpha -->
             <version>4.0.0-SNAPSHOT</version>
             <scope>test</scope>
         </dependency>
--- a/integration/test-integration/src/test/java/org/wso2am/micro/gw/tests/apim/APIMLifecycleBaseTest.java
+++ b/integration/test-integration/src/test/java/org/wso2am/micro/gw/tests/apim/APIMLifecycleBaseTest.java
@@ -1,0 +1,771 @@
+/*
+ * Copyright (c) 2021, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
+ *
+ * WSO2 Inc. licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.wso2am.micro.gw.tests.apim;
+
+import com.github.dockerjava.zerodep.shaded.org.apache.hc.core5.http.HttpStatus;
+import com.google.gson.Gson;
+import org.apache.commons.lang.StringUtils;
+import org.json.JSONArray;
+import org.json.JSONException;
+import org.json.JSONObject;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.wso2.am.integration.clients.publisher.api.ApiException;
+import org.wso2.am.integration.clients.publisher.api.v1.dto.APIDTO;
+import org.wso2.am.integration.clients.publisher.api.v1.dto.APIOperationsDTO;
+import org.wso2.am.integration.clients.store.api.v1.dto.ApplicationDTO;
+import org.wso2.am.integration.clients.store.api.v1.dto.ApplicationKeyDTO;
+import org.wso2.am.integration.clients.store.api.v1.dto.ApplicationKeyGenerateRequestDTO;
+import org.wso2.am.integration.test.impl.RestAPIPublisherImpl;
+import org.wso2.am.integration.test.impl.RestAPIStoreImpl;
+import org.wso2.am.integration.test.utils.APIManagerIntegrationTestException;
+import org.wso2.am.integration.test.utils.base.APIMIntegrationConstants;
+import org.wso2.am.integration.test.utils.bean.APILifeCycleAction;
+import org.wso2.am.integration.test.utils.bean.APILifeCycleState;
+import org.wso2.am.integration.test.utils.bean.APIRequest;
+import org.wso2.am.integration.test.utils.bean.APIRevisionDeployUndeployRequest;
+import org.wso2.am.integration.test.utils.bean.APIRevisionRequest;
+import org.wso2.carbon.automation.engine.context.beans.User;
+import org.wso2.carbon.automation.test.utils.http.client.HttpResponse;
+import org.wso2am.micro.gw.tests.common.model.API;
+import org.wso2am.micro.gw.tests.context.MicroGWTestException;
+import org.wso2am.micro.gw.tests.util.TestConstant;
+import org.wso2am.micro.gw.tests.util.Utils;
+
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import javax.ws.rs.core.Response;
+import javax.xml.xpath.XPathExpressionException;
+
+import static org.testng.Assert.assertEquals;
+
+public class APIMLifecycleBaseTest extends APIMWithMgwBaseTest {
+    private static final Logger LOGGER = LoggerFactory.getLogger(APIMLifecycleBaseTest.class);
+
+    /**
+     * Map of the predefined API requests. Key represents the api name and the value represents the APIRequest.
+     */
+    private static Map<String, APIRequest> apiRequestListByAPIName = new LinkedHashMap<>();
+
+    /**
+     * Map that binds the index of the api request to the api name.
+     */
+    private static Map<Integer, String> apiRequestListByIndex = new LinkedHashMap<>();
+
+    /**
+     * Map of the predefined application info. Key represents the application name and the value represents the
+     * Application instance.
+     */
+    private static Map<String, Application> applicationListByName = new LinkedHashMap<>();
+
+    /**
+     * Map that binds the index of the application to the application name.
+     */
+    private static Map<Integer, String> applicationListByIndex = new LinkedHashMap<>();
+
+    static {
+        // list of predefined applications.
+        addToApplicationList("SubscriptionValidationTestApp",
+                             new Application("SubscriptionValidationTestApp",
+                                             "Test Application for SubscriptionValidationTestCase",
+                                             TestConstant.APPLICATION_TIER.UNLIMITED,
+                                             ApplicationDTO.TokenTypeEnum.JWT));
+
+        // list of predefined api requests
+        String apiName = "SubscriptionValidationTestAPI";
+        String apiContext = "subscriptionValidationTestAPI";
+        String apiEndPointPostfixUrl = "/v2";
+
+        try {
+            APIRequest apiRequest = new APIRequest(apiName, apiContext,
+                                                   new URL(Utils.getDockerMockServiceURLHttp(apiEndPointPostfixUrl)));
+            String API_VERSION_1_0_0 = "1.0.0";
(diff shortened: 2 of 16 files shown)


integration/test-integration/pom.xml
```diff
@@ -44,6 +44,14 @@
                     <suiteXmlFiles>
                         <suiteXmlFile>src/test/resources/testng.xml</suiteXmlFile>
                     </suiteXmlFiles>
+                    <systemProperties>
+                        <property>
+                            <name>framework.resource.location</name>
+                            <value>
+                                ${basedir}/src/test/resources/
```
SuKSW, line 51: This location can be accessed using the path as [TestConstant.TEST_RESOURCES_PATH](https://github.com/wso2/product-microgateway/blob/9509397ca5e9a886815b8759c342e13a52867ba3/integration/test-integration/src/test/java/org/wso2am/micro/gw/tests/util/TestConstant.java#L75). An example can be found here: https://github.com/wso2/product-microgateway/blob/9509397ca5e9a886815b8759c342e13a52867ba3/integration/test-integration/src/test/java/org/wso2am/micro/gw/tests/util/ApictlUtils.java#L50

integration/test-integration/src/test/java/org/wso2am/micro/gw/tests/context/MgwServerInstance.java
```diff
@@ -118,68 +103,19 @@ public MgwServerInstance(String confPath, boolean tlsEnabled, boolean customJwtT
...
      * @throws MicroGWTestException
      */
-    public static void createTmpMgwSetup(boolean customJwtTransformerEnabled) throws IOException, MicroGWTestException {
-        File targetClassesDir = new File(MgwServerInstance.class.getProtectionDomain().getCodeSource().
-                getLocation().getPath());
-        String targetDir = targetClassesDir.getParentFile().toString();
-        final Properties properties = new Properties();
-        properties.load(MgwServerInstance.class.getClassLoader().getResourceAsStream("project.properties"));
-
-        Utils.copyDirectory(targetDir + File.separator + "micro-gwtmp" +  File.separator +
```
SuKSW, line 150: I think we need this section. Because when the default config gets replaced by a new config.toml and something like a jwt transformer gets added into the dropin folder, the folder that was originally extracted would change. So to get back the original mgw with the default behavior, we will need this section.

sajinieKavindya: I have created an abstract parent class and moved the commonly used code there. Please refer to MgwServerImpl.java line #51

integration/test-integration/src/test/java/org/wso2am/micro/gw/tests/context/APIManagerWithMgwServerInstance.java
```diff
@@ -0,0 +1,90 @@
...
+                TestConstant.APIM_SERVICE_NAME_IN_DOCKER_COMPOSE,
+                Wait.forLogMessage("\\/localhost:9443\\/carbon\\/", 1));
+    }
+
+    /**
+     * This will create a separate mgw setup in the target/server-tmp directory to execute the tests.
+     *
+     * @throws MicroGWTestException if something goes wring while copying the config file
+     */
+    private void createTmpMgwSetup() throws MicroGWTestException {
```
SuKSW, line 87: Shall we move this to the previous location because this is not APIM specific?

sajinieKavindya: This function is not removed from the previous location (MgwServerInstance.java). Just the method signature is changed there.

sajinieKavindya: This function is not removed from the previous location (MgwServerInstance.java). Just the method signature is changed there.

integration/test-integration/pom.xml
sajinieKavindya: This is just setting a system property called "framework.resource.location" (inside the pom.xml), so that the automation engine can configure the automation.xml before running the apim related integration tests.

integration/test-integration/src/test/java/org/wso2am/micro/gw/tests/testCases/subscription/SubscriptionValidationTestCase.java
```diff
@@ -0,0 +1,142 @@
...
+
+    private APIRequest apiRequest;
+    private String apiId;
+    private String applicationId;
+    private Map<String, String> requestHeaders;
+    private String endpointURL;
+
+    @BeforeClass(alwaysRun = true, description = "initialise the setup")
+    void setEnvironment() throws Exception {
+        super.init();
```
SuKSW, line 58: Is the super.init() specific to the SubscriptionValidationTestCase, if so shall we move it to somewhere other than the APIMLifecycleBaseTest class. If it is not specific, shall we move this line to APIManagerInitializeTestCase

integration/test-integration/src/test/java/org/wso2am/micro/gw/tests/apim/APIMLifecycleBaseTest.java
```diff
@@ -0,0 +1,771 @@
...
+
+    /**
+     * Retrieve  the value from JSON object bu using the key.
+     *
+     * @param httpResponse - Response that containing the JSON object in it response data.
+     * @param key          - key of the JSON value the need to retrieve.
+     * @return String - The value of provided key as a String
+     * @throws MicroGWTestException - Exception throws when resolving the JSON object in the HTTP response
+     */
+    protected String getValueFromJSON(HttpResponse httpResponse, String key) throws MicroGWTestException {
```
SuKSW, line 552: Shall we move methods similar to these to a suitable file inside the utils folder and make them static?

integration/test-integration/src/test/java/org/wso2am/micro/gw/tests/testCases/subscription/SubscriptionValidationTestCase.java
sajinieKavindya: This is not specific only to SubscriptionValidationTestCase. All the apim related test cases need to call the super.init() function to initialize the context required for that particular test case. This call needs to be done in each such test cases as some test cases might need to initialize the context for a particular user. For that purpose, they can use `void init(TestUserMode userMode)`.

Also, this method cannot be moved to APIManagerInitializeTestCase since there is no way we can maintain the initialized variables throughout the test suite.

integration/test-integration/src/test/java/org/wso2am/micro/gw/tests/apim/APIMLifecycleBaseTest.java
sajinieKavindya: +1

integration/test-integration/src/test/java/org/wso2am/micro/gw/tests/apim/APIMLifecycleBaseTest.java
```diff
@@ -0,0 +1,771 @@
...
+     * Publish an API using REST.
+     *
+     * @param apiId                   - UUID of the API,
+     * @param publisherRestClient     - Instance of APIPublisherRestClient
+     * @param isRequireReSubscription - If publish with re-subscription required option true else false.
+     * @return HttpResponse - Response of the API Publishing activity
+     * @throws MicroGWTestException -  Exception throws by the method call of changeAPILifeCycleStatusToPublish() in
+     *                              APIPublisherRestClient.java.
+     */
+    protected HttpResponse publishAPI(String apiId, RestAPIPublisherImpl publisherRestClient,
```
SuKSW, line 134: If possible, shall we move these methods to a seperate file so that these methods can be used by a different test case that does not want to extend APIMLifecycleBaseTest?

integration/test-integration/src/test/java/org/wso2am/micro/gw/tests/apim/APIMWithMgwBaseTest.java
```diff
@@ -0,0 +1,336 @@
...
+                LOGGER.info("Data: " + response.getData());
+                if (response.getData().contains(expectedResponse)) {
+                    LOGGER.info("API :" + apiName + " with version: " + apiVersion + " with expected response "
+                                        + expectedResponse + " found");
+                    break;
+                } else {
+                    try {
+                        Thread.sleep(500);
+                    } catch (InterruptedException ignored) {
+                        // do nothing
```
SuKSW, line 241:
```suggestion
                        Thread.currentThread.interrupt()
```
I think we are supposed to call this here.

integration/test-integration/src/test/java/org/wso2am/micro/gw/tests/context/APIManagerWithMgwServerInstance.java
SuKSW: Sorry, I meant the content below it. Edited my comment above to suit the scenario.

integration/test-integration/src/test/java/org/wso2am/micro/gw/tests/testCases/subscription/SubscriptionValidationTestCase.java
SuKSW: If there can be some test cases might need to initialize the context for a particular user then it shouldn't be common to the suit right?

integration/test-integration/src/test/java/org/wso2am/micro/gw/tests/context/APIManagerWithMgwServerInstance.java
SuKSW: The content where it copies the extracted mgw to a temp location.

integration/test-integration/src/test/java/org/wso2am/micro/gw/tests/testCases/subscription/SubscriptionValidationTestCase.java
SuKSW: The reason to ask this is because both the subscription related test case and the initialization test case extends the same APIMLifecycleBaseTest class. Was thinking whether the two stages can be separated into two classes because they'd anyway be two different instances of the same super class.

integration/test-integration/src/test/java/org/wso2am/micro/gw/tests/apim/APIMLifecycleBaseTest.java
sajinieKavindya: This APIMLifecycleBaseTest class is like a utility class written to the apim related test cases. if we move these methods, we need to move 95% of the code to another class. And these methods are tightly coupled to APIM scenarios. So, if someone needs to use the method, I don't see any issue extending this class (even with a class name like this). 
So, WDYT?

integration/test-integration/src/test/java/org/wso2am/micro/gw/tests/apim/APIMWithMgwBaseTest.java
sajinieKavindya: +1

integration/test-integration/src/test/java/org/wso2am/micro/gw/tests/testCaseBefore/APIManagerInitializeTestCase.java
```diff
@@ -0,0 +1,123 @@
...
+import java.net.URL;
+import java.util.Objects;
+import java.util.concurrent.Callable;
+import java.util.concurrent.TimeUnit;
+
+public class APIManagerInitializeTestCase extends APIMLifecycleBaseTest {
+
+    @BeforeTest(description = "start the mgw along with apim server")
+    void start() throws Exception {
+        URL configToml = getClass().getClassLoader().getResource("apim/config.toml");
```
SuKSW, line 47: Let's use path.separater instread of direclty using  /

integration/test-integration/pom.xml
SuKSW: Got it. Thanks. +1

integration/test-integration/src/test/java/org/wso2am/micro/gw/tests/context/APIManagerWithMgwServerInstance.java
sajinieKavindya: +1

integration/test-integration/src/test/java/org/wso2am/micro/gw/tests/testCaseBefore/APIManagerInitializeTestCase.java
sajinieKavindya: `getClass().getClassLoader().getResource()` - this method is a platform independent. if you see the javadoc of the getResource() method, it clearly mentions that the path should be separated by "/". So, why use the path.separater?

integration/test-integration/src/test/java/org/wso2am/micro/gw/tests/testCases/subscription/SubscriptionValidationTestCase.java
sajinieKavindya: > If there can be some test cases might need to initialize the context for a particular user then it shouldn't be common to the suit right?

I don't think we need to categorize test suits based on the user the test case is going to use. Then, there will be unnecessary suits created just for different users

SuKSW: Got it. Let's leave it as it is.

integration/test-integration/src/test/java/org/wso2am/micro/gw/tests/apim/APIMLifecycleBaseTest.java
SuKSW: The reason to make the suggestion was because they are utility methods rather than part of the actual best test case. Wouldn't it be difficult to identify the necessary parts of the best test case if we also include the util methods  in it?

integration/test-integration/src/test/java/org/wso2am/micro/gw/tests/testCaseBefore/APIManagerInitializeTestCase.java
SuKSW: Just learnt what the getResource() method is about. Thanks. +1

integration/test-integration/pom.xml
```diff
@@ -220,6 +228,18 @@
             <artifactId>org.wso2.micro.gateway.enforcer</artifactId>
             <version>4.0.0-m10-SNAPSHOT</version>
         </dependency>
+        <dependency>
+            <groupId>org.wso2.am</groupId>
+            <artifactId>org.wso2.am.integration.common.test.utils</artifactId>
+            <version>4.0.0-SNAPSHOT</version>
```
Rajith90, line 234: Shall we use something like 4.0.0-alpha

sajinieKavindya: +1

sajinieKavindya: 4.0.0-alpha has not been released to nexus yet. Therefore, I will leave here as it is for the moment. Added a todo comment.

Rajith90 (approved):

PR #1884, opened 2021-04-10
Bring in java SPI to resolve customJWTTransformer

### LINKED ISSUE #1885: Resolve custom implementations using JAVA SPI Interface
### Describe your problem(s)
Currently, the custom implementations for JWTGenerator, JWTTransformer is loaded by iterating and traversing over JAR files. But the most suitable option would be using JAVA SPI or Class.forName method where necessary.

For JWT Transformer Implementation, the user is not supposed to provide a className. The corresponding issuer, whether the transformer needs to enabled information are fetched from annotations. Hence JAVA SPI is used to provide/resolve that.

For Custom JWT Generator Implemetation, the user should provide the className. In that case, class.forName method should be used.

### How will you implement it

---

VirajSalaka:
### Purpose
$subject

### Issues

Fixes #1885

### Automation tests
 - Unit tests added: No
 - Integration tests added: Covered from existing tests

### Tested environments

Not Tested

---
- [x] Assigned 'Type' label
- [x] Assigned the project

### DIFF
--- a/integration/test-integration/src/main/resources/META-INF/services/org.wso2.carbon.apimgt.common.gateway.jwttransformer.JWTTransformer
+++ b/integration/test-integration/src/main/resources/META-INF/services/org.wso2.carbon.apimgt.common.gateway.jwttransformer.JWTTransformer
@@ -1,0 +1,1 @@
+org.wso2.choreo.connect.tests.CustomJWTTransformer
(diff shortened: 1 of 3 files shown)


integration/test-integration/src/main/resources/META-INF/services/org.wso2.carbon.apimgt.common.gateway.jwttransformer.JWTTransformer
```diff
@@ -0,0 +1 @@
+org.wso2.choreo.connect.tests.CustomJWTTransformer
```
Rajith90, line 1: Do we need a jar to be committed in the repo

VirajSalaka: not committed any where. Rather user needs to add it to dropins directory.
When creating the custom impl jar, the user needs to mention his class under  `META-INF/services/org.wso2.carbon.apimgt.common.gateway.jwttransformer.JWTTransformer` file

Rajith90 (approved):

PR #1982, opened 2021-04-29
Update docker image and apim deployment.toml

menakaj:
* Change the ws and wss endpoint ports
* Modify the adapter Dockerfile to upgrade the alpine os

### Issues

Fixes https://github.com/wso2/product-microgateway/issues/1983

### Automation tests
 - Unit tests added: Yes/No

### Tested environments

Not Tested

---
- [x] Assigned 'Type' label
- [x] Assigned the project


NipunaPrashan (approved):
LGTM

Rajith90 (approved):

adapter/src/main/resources/Dockerfile
```diff
@@ -15,7 +15,8 @@
 # -----------------------------------------------------------------------
 
 FROM alpine:3.13.4
-LABEL maintainer="WSO2 Docker Maintainers <dev@wso2.org>" 
+LABEL maintainer="WSO2 Docker Maintainers <dev@wso2.org>"
+RUN apk update && apk upgrade --no-cache
```
pubudu538, line 19: Without this cant we use alpine:3.13.5? Is the vulnerability still there in 3.13.5?

menakaj:
```
2021-05-03T12:12:37.479+0530	WARN	This OS version is no longer supported by the distribution: alpine 3.13.5
2021-05-03T12:12:37.479+0530	WARN	The vulnerability detection may be insufficient because security updates are not provided
```

Trivy does not support 13.5 yet. 

+1 for updating to the latest version. However, we will have to update the baseimage in our dockerfiles everytime a new vulnerability is identified. 
With apk upgrade, it will be automatically fixed if there is a fix released to that version.
Wdyt?

pubudu538: Every time we build a new docker, it creates the latest image and we cannot guarantee any os version here. As long as the image works, we are fine. We can run a test suite for this image and verify functionality.

menakaj: `apk upgrade` does not update the os version and only updates the fixed packages.

Issue #2041, opened 2021-05-31
Introduce per tenant/ per organization resource maps


menakaj:
### Describe your problem(s)
Currently in Choreo Connect, the api resources are stored in maps against vHost:apiName:apiVersion as the key. This cause issues when apis with same name and version are deployed. 

### How will you implement it

---
#### Related Issues:

https://github.com/wso2-enterprise/choreo/issues/4273

PR #2042, opened 2021-05-31
Introduce tenant domain/ organizationId based entity maps

menakaj:
### Purpose
The internal maps which are being used in Choreo Connect use, VHost:apiName:apiVersion as the map key. This have become an issue when multiple organizations having apis with the same name and version.

To fix this issue, this PR introduces, organizationID/ tenantId as the map key. 

Methodology:
1. The tenant domain parameter in API-M event is used as the organization Id to store the resources
2. In API-CTL scenario, the OrganizationId can be provided in the api.yaml file.
3. If the orgId is not provided (either in apictl mode or with api-manager) the authenticated user's tenant domain will be set as the organizationId.

### Issues

Fixes https://github.com/wso2/product-microgateway/issues/2041

### Automation tests
 - Unit tests added: Yes/No

### Tested environments

Not Tested

---
- [x] Assigned 'Type' label
- [x] Assigned the project


Rajith90:
Could you please Update the docs accordingly as well

adapter/internal/discovery/xds/server.go
```diff
@@ -882,129 +930,129 @@ func UpdateEnforcerThrottleData(throttleData *throttle.ThrottleData) {
...
-			// Successful message, set the current state of the enforcer apis to the state cache.
-			buildAndStoreSuccessState(requestEvent.Node, requestEvent.Version)
-
-			// Generate the router resources and update the router.
-			listeners, clusters, routes, endpoints, _ := GenerateEnvoyResoucesForLabel(requestEvent.Node)
-			UpdateXdsCacheWithLock(requestEvent.Node, endpoints, clusters, routes, listeners)
-		}
-	}
-}
+// // buildAndStoreSuccessState method builds the latest successful enforcer and router resources and returns as
```
praminda, line 933: why commented?

menakaj: This code is not being executed. Will remove this

adapter/config/types.go
```diff
@@ -355,6 +355,7 @@ type controlPlane struct {
 	EnvironmentLabels       []string      `toml:"environmentLabels"`
 	RetryInterval           time.Duration `toml:"retryInterval"`
 	SkipSSLVerification     bool          `toml:"skipSSLVerification"`
+	OperationMode           string        `toml:"operationMode"`
```
Rajith90, line 358: Lets remove this

adapter/internal/discovery/xds/consul.go
```diff
@@ -35,14 +35,14 @@ const (
 	transportSocketName = "envoy.transport_sockets.tls"
 )
 
-func startConsulServiceDiscovery() {
-	for apiKey, clusterList := range openAPIClustersMap {
+func startConsulServiceDiscovery(organizationID string) {
```
Rajith90, line 38: Where is this method called, We should call with org Id since the method signature has changed right?

Rajith90 (changes requested):

adapter/config/types.go
menakaj: Removed

adapter/internal/discovery/xds/consul.go
menakaj: Yes. This is being called when the api is added. https://github.com/wso2/product-microgateway/pull/2042/files#diff-0bb37b8dc28a8ae4fbd9179a4b374745a9868b20fa545baa063548dfe63294b6R342

praminda (approved):

Rajith90 (approved):

PR #2076, opened 2021-06-16
Remove redundant log config read operation.

### LINKED ISSUE #2077: LogConfig and MGW_HOME read operations are repeated.
### Describe your problem(s)

As per current implementation, the log_config.toml read operation and GetMGWHome() operation functionality has been repeated twice despite they share the same functionality.

One implementation resides in pkg/config directory while the other implementation can be seen in config directory.

### Describe your solution

Following facts should be taken into account while solving the issue.

- The log_config.toml parsing operation should reside within the pkg directory as the loggers are initialized within both pkg and internal directory. 
- The same implementation should be reused within global adapter to read and parse logs.
The log_config.toml 's file name and directory may change when it is reused within a global adapter (not necessarily located within `conf/log_config.toml` relative path)

Based on these, 

- a new environmental variable is introduced `LOG_CONFIG_PATH`. And this has the top priority for resolving log_config.toml's file path.
- if it is not available, relative path based approach (current implementation) will be used. (Therefore this preserves backward compatibility)

### How will you implement it

---

VirajSalaka:
### Purpose

The log_config toml parsing related implementation will be reused by the both global adapter and the adapter implementation. Hence it has already been moved to the pkg directory. 

This PR would avoid the following redundant operations introduced within the fore-mentioned case.

- GetMGWHome - which retrieves the MGW_HOME environmental variable
- ReadLogConfigs - which parses the log_config.toml

### Issues

Fixes #2077 

### Automation tests
 - Unit tests added: No
 - Integration tests added: No

### Tested environments

Not Tested

---
- [ ] Assigned 'Type' label
- [ ] Assigned the project


adapter/pkg/config/parser.go
```diff
@@ -58,6 +62,23 @@ func GetMgwHome() string {
...
+func getLogConfigPath() string {
+	logConfigPath = os.Getenv(logConfigPathEnvVariable)
+	if len(strings.TrimSpace(logConfigPath)) == 0 {
+		//for backward compatibility
+		logConfigPath = GetMgwHome() + relativeLogConfigPath
+	}
+	_, err := os.Stat(logConfigPath)
+	if err != nil {
+		logger.Fatal("Log configuration file not found.", err)
+		panic(err)
```
VirajSalaka, line 77: No need to panic if the the log_config.toml is not available
And log level should be set to Warn

adapter/pkg/config/parser.go
```diff
@@ -69,15 +90,10 @@ func GetMgwHome() string {
...
-		if err != nil {
-			logger.Fatal("Log configuration file not found.", err)
-			panic(err)
-		}
-		content, readErr := ioutil.ReadFile(mgwHome + relativeLogConfigPath)
+		content, readErr := ioutil.ReadFile(getLogConfigPath())
 		if readErr != nil {
 			logger.Fatal("Error reading log configurations. ", readErr)
-			panic(err)
+			panic(readErr)
```
VirajSalaka, line 96: No need to panic

adapter/config/parser.go
```diff
@@ -31,17 +31,14 @@ import (
 
 	toml "github.com/pelletier/go-toml"
 	logger "github.com/sirupsen/logrus"
+	pkg_conf "github.com/wso2/product-microgateway/adapter/pkg/config"
```
VirajSalaka, line 34: pkgconf

adapter/internal/oasparser/envoyconf/envoyconf_internal_test.go
```diff
@@ -112,12 +112,18 @@ func TestCreateRoute(t *testing.T) {
 		Enabled:     &wrappers.BoolValue{Value: false},
 	}}
 
+	TimeOutConfig := ptypes.DurationProto(60 * time.Second)
+
+	IdleTimeOutConfig := ptypes.DurationProto(300 * time.Second)
```
mpmunasinghe, line 117: what is the 300 value here ?

adapter/pkg/config/parser.go
```diff
@@ -66,25 +88,28 @@ func GetMgwHome() string {
 // from where the executable is called from.
 //
 // Returns the log configuration object mapped from the configuration file during the startup.
-func ReadLogConfigs() (*LogConfig, error) {
+func ReadLogConfigs() *LogConfig {
 	onceLogConfigRead.Do(func() {
-		adapterLogConfig = new(LogConfig)
-		_, err := os.Stat(GetMgwHome() + relativeLogConfigPath)
+		adapterLogConfig = getDefaultLogConfig()
```
mpmunasinghe, line 93: Couldn't find the local function for getDefaultLogConfig()

VirajSalaka: The function is moved to log_types.go file. Since both files are located within the same package, it is resolved.

VirajSalaka:
Closing the PR since it contains additional commits other than intended changes.

PR #2082, opened 2021-06-21
Handle malformed endpoints

### LINKED ISSUE #2083: Adapter crashes when malformed endpoint is provided
### Description:
Choreo Connect adapter crashes when trying to parse malformed url and does not recover.

- OS:
- Client:
- Env (Docker/K8s):

---

ashera96:
### Purpose
Handle malformed endpoints

### Issues
Fixes #2083 

### Automation tests
 - Unit tests added: Yes/No

### Tested environments

Not Tested

---
- [x] Assigned 'Type' label
- [x] Assigned the project

### DIFF
--- a/adapter/internal/oasparser/model/open_api.go
+++ b/adapter/internal/oasparser/model/open_api.go
@@ -19,6 +19,7 @@ package model
 
 import (
 	"encoding/json"
+	"errors"
 	"net/url"
 	"strconv"
 	"strings"
@@ -37,7 +38,8 @@ import (
 // UUID.
 //
 // No operation specific information is extracted.
-func (swagger *MgwSwagger) SetInfoOpenAPI(swagger3 openapi3.Swagger) {
+func (swagger *MgwSwagger) SetInfoOpenAPI(swagger3 openapi3.Swagger) error {
+	var err error
 	if swagger3.Info != nil {
 		swagger.description = swagger3.Info.Description
 		swagger.title = swagger3.Info.Title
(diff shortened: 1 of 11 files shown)


adapter/internal/oasparser/model/open_api.go
```diff
@@ -105,14 +116,18 @@ func setResourcesOpenAPI(openAPI openapi3.Swagger) []Resource {
 						continue
 					}
 					endpoint := getHostandBasepathandPort(serverEntry.URL)
-					resource.productionUrls = append(resource.productionUrls, endpoint)
+					if endpoint != nil {
+						resource.productionUrls = append(resource.productionUrls, *endpoint)
+					} else {
```
VirajSalaka, line 121: if the endpoint is not an error right? rather than not a nil

ashera96: Based on the changes made, the function `getHostandBasepathandPort` either returns a pointer to the endpoint or a nil. Hence, it will not return an error.

pubudu538 (approved):

VirajSalaka (approved):

adapter/internal/oasparser/operator/operator.go
```diff
@@ -51,7 +51,10 @@ func GetMgwSwagger(apiContent []byte) model.MgwSwagger {
 		if err != nil {
 			logger.LoggerOasparser.Error("Error openAPI unmarsheliing", err)
```
GDLMadushanka, line 52: is it not required to return this unmarshal err as well?

ashera96: Just a log would suffice in this particular instance as we just need to know that an unmarshalling error occurred. The problem with returning at this point is, this will stop applying the API as a whole. But since it won't be a breaking error for basic functionalities of the API, we decided to not return. @VirajSalaka correct me if I'm wrong.

PR #2100, opened 2021-06-25
Remove redundant log config read operation

### LINKED ISSUE #2077: LogConfig and MGW_HOME read operations are repeated.

VirajSalaka:
### Purpose

The log_config toml parsing related implementation will be reused by the both global adapter and the adapter implementation. Hence it has already been moved to the pkg directory. 

This PR would avoid the following redundant operations introduced within the fore-mentioned case.

- GetMGWHome - which retrieves the MGW_HOME environmental variable
- ReadLogConfigs - which parses the log_config.toml

### Issues

Fixes #2077 

### Automation tests
 - Unit tests added: No
 - Integration tests added: No

### Tested environments

Docker-compose setup

---
- [x] Assigned 'Type' label
- [x] Assigned the project


VirajSalaka:
Review comments can be viewed here
https://github.com/wso2/product-microgateway/pull/2076

adapter/pkg/config/parser.go
```diff
@@ -66,25 +87,27 @@ func GetMgwHome() string {
...
+		adapterLogConfig = getDefaultLogConfig()
+		filePath, err := GetLogConfigPath()
+		if err == nil {
+			content, readErr := ioutil.ReadFile(filePath)
+			if readErr != nil {
+				logger.Fatal("Proceeding with default log configuration as error occured while reading log configurations ", readErr)
+			} else {
+				parseErr := toml.Unmarshal(content, adapterLogConfig)
+				if parseErr != nil {
+					logger.Fatal("Proceeding with default log configuration as error occured while parsing the log configuration ", parseErr)
```
mpmunasinghe, line 101: Fatal will exit the process without continuing with the default configs?

mpmunasinghe (approved):

adapter/pkg/config/parser.go
VirajSalaka: resolved via https://github.com/wso2/product-microgateway/pull/2100/commits/d40f354903cd9fae24039f4a8a123f162a5930a2

adapter/config/parser.go
```diff
@@ -77,15 +74,15 @@ const (
...
 func ReadConfigs() (*Config, error) {
 	onceConfigRead.Do(func() {
 		adapterConfig = defaultConfig
-		_, err := os.Stat(GetMgwHome() + relativeConfigPath)
+		_, err := os.Stat(pkgconf.GetMgwHome() + relativeConfigPath)
 		if err != nil {
 			logger.Fatal("Configuration file not found.", err)
 		}
-		content, readErr := ioutil.ReadFile(mgwHome + relativeConfigPath)
+		content, readErr := ioutil.ReadFile(pkgconf.GetMgwHome() + relativeConfigPath)
```
Rajith90, line 85: can't we `config` as the package name instead of `pkgconf`

adapter/internal/adapter/adapter.go
```diff
@@ -158,12 +158,14 @@ func Run(conf *config.Config) {
...
 	watcherLogConf, _ := fsnotify.NewWatcher()
-	errC := watcherLogConf.Add(config.GetMgwHome() + "/conf/log_config.toml")
+	logConfigPath, errC := config.GetLogConfigPath()
+	if errC == nil {
+		errC = watcherLogConf.Add(logConfigPath)
+	}
 
 	if errC != nil {
-		logger.LoggerMgw.Fatal("Error reading the log configs. ", errC)
+		logger.LoggerMgw.Error("Error reading the log configs. ", errC)
```
Rajith90, line 168: This error detected for dynamic updates of the configs , isn't it? otherwise failing to read the config should kill the adapter

VirajSalaka:
```
adapter_1   | 2021-06-30 05:55:33 ERRO [parser.go:101] - [config.ReadLogConfigs.func1] [-] Proceeding with default log configuration as error occured while parsing the log configuration (26, 12): unescaped control character U+000A
```
It would print something like this and proceed.

this is for dynamic update of Log configs, not configs

adapter/config/parser.go
VirajSalaka: this one's package name is config. the reference is pkgconf
```
pkgconf "github.com/wso2/product-microgateway/adapter/pkg/config"
```

so I would rename the reference as configCommon

Rajith90 (approved):

choreo-cicd:
[failed] dev-deployment : 20210825.10

choreo-cicd:
[succeeded] prod-deployment : 20210831.1

PR #2103, opened 2021-06-28
Handle adapter crash issue when malformed endpoints are provided

ashera96:
### Purpose
This PR fixes the adapter crash issue when a malformed endpoint is encountered.
Note: Extension on the temp fix (#2082).

### Issues
Fixes: [#5571](https://github.com/wso2-enterprise/choreo/issues/5571)

### Automation tests
 - Unit tests added: Yes

### Tested environments

Not Tested

---
- [x] Assigned 'Type' label
- [x] Assigned the project

### DIFF
--- a/adapter/internal/adapter/adapter.go
+++ b/adapter/internal/adapter/adapter.go
@@ -312,7 +312,7 @@ func fetchAPIsOnStartUp(conf *config.Config) {
 				logger.LoggerMgw.Debugf("Time Duration for retrying: %v", conf.ControlPlane.RetryInterval*time.Second)
 				time.Sleep(conf.ControlPlane.RetryInterval * time.Second)
 				logger.LoggerMgw.Infof("Retrying to fetch API data from control plane.")
-				go synchronizer.FetchAPIs(&d.APIUUID, d.GatewayLabels, c)
+				synchronizer.FetchAPIs(&d.APIUUID, d.GatewayLabels, c)
 			}(data)
 		}
 	}
--- a/adapter/internal/adapter/adapter.go
+++ b/adapter/internal/adapter/adapter.go
@@ -312,7 +312,7 @@ func fetchAPIsOnStartUp(conf *config.Config) {
 				logger.LoggerMgw.Debugf("Time Duration for retrying: %v", conf.ControlPlane.RetryInterval*time.Second)
 				time.Sleep(conf.ControlPlane.RetryInterval * time.Second)
 				logger.LoggerMgw.Infof("Retrying to fetch API data from control plane.")
-				synchronizer.FetchAPIs(&d.APIUUID, d.GatewayLabels, c)
+				go synchronizer.FetchAPIs(&d.APIUUID, d.GatewayLabels, c)
 			}(data)
 		}
 	}
--- a/adapter/internal/api/apis_impl.go
+++ b/adapter/internal/api/apis_impl.go
@@ -25,7 +25,7 @@ import (
 	"encoding/hex"
 	"encoding/json"
 	"errors"
-	"github.com/wso2/product-microgateway/adapter/internal/notifier"
+	"fmt"
 	"io/ioutil"
 	"os"
 	"strings"
@@ -34,6 +34,7 @@ import (
 	apiModel "github.com/wso2/product-microgateway/adapter/internal/api/models"
 	xds "github.com/wso2/product-microgateway/adapter/internal/discovery/xds"
 	"github.com/wso2/product-microgateway/adapter/internal/loggers"
+	"github.com/wso2/product-microgateway/adapter/internal/notifier"
 	mgw "github.com/wso2/product-microgateway/adapter/internal/oasparser/model"
 	"github.com/wso2/product-microgateway/adapter/internal/oasparser/utills"
 	"github.com/wso2/product-microgateway/adapter/pkg/tlsutils"
@@ -231,8 +232,7 @@ func verifyMandatoryFields(apiJSON config.APIJsonData) error {
 
 // ApplyAPIProjectFromAPIM accepts an apictl project (as a byte array), list of vhosts with respective environments
 // and updates the xds servers based upon the content.
-func ApplyAPIProjectFromAPIM(payload []byte, vhostToEnvsMap map[string][]string) ([]*notifier.DeployedAPIRevision,error) {
-	var deployedRevisionList []*notifier.DeployedAPIRevision
+func ApplyAPIProjectFromAPIM(payload []byte, vhostToEnvsMap map[string][]string) (deployedRevisionList []*notifier.DeployedAPIRevision, err error) {
 	apiProject, err := extractAPIProject(payload)
 	if err != nil {
 		return deployedRevisionList, err
@@ -242,6 +242,14 @@ func ApplyAPIProjectFromAPIM(payload []byte, vhostToEnvsMap map[string][]string)
 		return deployedRevisionList, err
 	}
 
+	// handle panic
+	defer func() {
+		if r := recover(); r != nil {
+			loggers.LoggerAPI.Error("Recovered from panic. ", r)
+			err = fmt.Errorf("%v:%v with UUID \"%v\"", apiInfo.Name, apiInfo.Version, apiInfo.ID)
+		}
+	}()
+	
 	if apiProject.OrganizationID == "" {
 		apiProject.OrganizationID = config.GetControlPlaneConnectedTenantDomain()
 	}
@@ -294,7 +302,7 @@ func ApplyAPIProjectFromAPIM(payload []byte, vhostToEnvsMap map[string][]string)
 
 // ApplyAPIProjectInStandaloneMode is called by the rest implementation to differentiate
 // between create and update using the override param
-func ApplyAPIProjectInStandaloneMode(payload []byte, override *bool) error {
+func ApplyAPIProjectInStandaloneMode(payload []byte, override *bool) (err error) {
 	apiProject, err := extractAPIProject(payload)
 	apiProject.OrganizationID = config.GetControlPlaneConnectedTenantDomain()
 	if err != nil {
@@ -304,6 +312,15 @@ func ApplyAPIProjectInStandaloneMode(payload []byte, override *bool) error {
 	if err != nil {
 		return err
 	}
+
+	// handle panic
+	defer func() {
+		if r := recover(); r != nil {
+			loggers.LoggerAPI.Error("Recovered from panic. ", r)
+			err = fmt.Errorf("%v:%v with UUID \"%v\"", apiInfo.Name, apiInfo.Version, apiInfo.ID)
+		}
+	}()
+
 	// TODO (renuka) when len of apiProject.deployments is 0, return err "nothing deployed" <- check
 	var overrideValue bool
 	if override == nil {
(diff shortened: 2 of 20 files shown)


adapter/internal/adapter/adapter.go
```diff
@@ -308,7 +308,7 @@ func fetchAPIsOnStartUp(conf *config.Config) {
 				logger.LoggerMgw.Debugf("Time Duration for retrying: %v", conf.ControlPlane.RetryInterval*time.Second)
 				time.Sleep(conf.ControlPlane.RetryInterval * time.Second)
 				logger.LoggerMgw.Infof("Retrying to fetch API data from control plane.")
-				synchronizer.FetchAPIs(&d.APIUUID, d.GatewayLabels, c)
+				go synchronizer.FetchAPIs(&d.APIUUID, d.GatewayLabels, c)
```
Rajith90, line 311: Ready state is set from this function, please check the readiness probes are working in k8s without any issue

adapter/internal/oasparser/model/mgw_swagger.go
```diff
@@ -244,18 +256,20 @@ func (swagger *MgwSwagger) SetXWso2ProductionEndpointMgwSwagger(productionURL st
...
 
-func (swagger *MgwSwagger) setXWso2PrdoductionEndpoint() error {
-	xWso2APIEndpoints := getXWso2Endpoints(swagger.vendorExtensions, productionEndpoints)
+func (swagger *MgwSwagger) setXWso2ProductionEndpoint() error {
+	xWso2APIEndpoints, err := getXWso2Endpoints(swagger.vendorExtensions, productionEndpoints)
 	if xWso2APIEndpoints != nil && len(xWso2APIEndpoints) > 0 {
 		swagger.productionUrls = xWso2APIEndpoints
-	} else {
+	} else if err != nil {
 		return errors.New("error encountered when extracting endpoints")
```
Rajith90, line 264: Better of we can provide more details about the endpoint url and etc

adapter/internal/oasparser/model/mgw_swagger.go
```diff
@@ -264,17 +278,19 @@ func (swagger *MgwSwagger) setXWso2PrdoductionEndpoint() error {
 }
 
 func (swagger *MgwSwagger) setXWso2SandboxEndpoint() error {
-	xWso2APIEndpoints := getXWso2Endpoints(swagger.vendorExtensions, sandboxEndpoints)
+	xWso2APIEndpoints, err := getXWso2Endpoints(swagger.vendorExtensions, sandboxEndpoints)
 	if xWso2APIEndpoints != nil && len(xWso2APIEndpoints) > 0 {
 		swagger.sandboxUrls = xWso2APIEndpoints
-	} else {
+	} else if err != nil {
 		return errors.New("error encountered when extracting endpoints")
```
Rajith90, line 285: Better if we can provide more details about the endpoint url and etc

adapter/internal/oasparser/model/mgw_swagger.go
```diff
@@ -244,18 +256,20 @@ func (swagger *MgwSwagger) SetXWso2ProductionEndpointMgwSwagger(productionURL st
...
 	}
 
 	//resources
 	for i, resource := range swagger.resources {
-		xwso2ResourceEndpoints := getXWso2Endpoints(resource.vendorExtensions, productionEndpoints)
-		if xwso2ResourceEndpoints != nil {
+		xwso2ResourceEndpoints, err := getXWso2Endpoints(resource.vendorExtensions, productionEndpoints)
+		if err != nil {
+			return err
```
VirajSalaka, line 271: Shall we log the error perform a continue here?
@Rajith90 wdyt?

adapter/internal/oasparser/model/mgw_swagger.go
VirajSalaka: If we propagate the error generated from `getHostNamePortBasepath` method it contains those information right? (rather than generating a new error)

adapter/internal/adapter/adapter.go
VirajSalaka: Could you please explain why we need a go-routine here?

adapter/internal/synchronizer/apis_fetcher.go
```diff
@@ -318,7 +318,7 @@ func FetchAPIsFromControlPlane(updatedAPIID string, updatedEnvs []string) {
 				logger.LoggerSync.Debugf("Time Duration for retrying: %v", conf.ControlPlane.RetryInterval*time.Second)
 				time.Sleep(conf.ControlPlane.RetryInterval * time.Second)
 				logger.LoggerSync.Info("Retrying to fetch API data from control plane.")
-				FetchAPIs(&updatedAPIID, finalEnvs, c)
+				go FetchAPIs(&updatedAPIID, finalEnvs, c)
```
VirajSalaka, line 321: why do we need a go-routine here?

adapter/internal/adapter/adapter.go
Rajith90: During the startup if this happens in the main thread, it can cause adapter to crash, isn't it. So we thought of taking this out to separate routine so the adapter will not crash if error during pulling the APIs. Only the ready state will not be set.

adapter/internal/oasparser/model/mgw_swagger.go
ashera96: The `getHostPortandBasePath` method logs the endpoint that causes the issue. This is only an intermediate step that is used to propagate the error. Which API in particular, is logged in the parent call (i.e. api name, version, id, etc.). Parent call as in `ApplyAPIProjectFromAPIM` or `ApplyAPIProjectInStandaloneMode`. This particular error is not logged anywhere. Do we still need to include all those details?

Please refer to the following adapter log when a malformed endpoint is used.

APIM_AdapterCrashFix.

adapter/internal/oasparser/model/mgw_swagger.go
ashera96: When it comes to resources, is it okay to continue with the deployment even after we detect a malformed endpoint?

adapter/internal/synchronizer/apis_fetcher.go
ashera96: Idea was to use a goroutine to make sure a single point of failure don't end up crashing the adapter. But since this call is already under a goroutine, we could remove it. @Rajith90 based on this logic, can we remove this?

adapter/internal/adapter/adapter.go
ashera96: Here too, this goroutine is already under an existing goroutine. So, it won't end up crashing. The root cause for the crash that we observed was the `Fatal` log which essentially does an os.Exit(1). I think we can remove this too.

Rajith90: If at the root level we use  routine then here we don't need I think

adapter/internal/synchronizer/apis_fetcher.go
Rajith90: We need a one go routine at the root level, otherwise we can't use the panic recover  function without a go routine AFAIU

adapter/internal/adapter/adapter.go
ashera96: Will remove this goroutine in that case

adapter/internal/oasparser/model/mgw_swagger.go
ashera96: We already log those specific details. What exact endpoint is logged at the `getHostPortandBasePath` function. When notifying that this API is no longer processed due to an error, we log the API related details at the root level. This is added just to propagate the error.

adapter/internal/synchronizer/apis_fetcher.go
ashera96: Verified by replicating a panic scenario. We can still recover successfully even after removing this goroutine. Hence removed this as well. Refer [218fe21](https://github.com/wso2/product-microgateway/pull/2103/commits/218fe21463db20eee3387a5310646aaa45a7d860)

VirajSalaka (approved):

menakaj (approved):

choreo-cicd:
[succeeded] dev-deployment : 20210727.4

choreo-cicd:
[succeeded] stage-deployment : 20210727.2

choreo-cicd:
[succeeded] prod-deployment : 20210729.1

adapter/internal/api/apis_impl.go
```diff
@@ -272,9 +280,9 @@ func ApplyAPIProjectFromAPIM(payload []byte, vhostToEnvsMap map[string][]string)
 		// first update the API for vhost
 		deployedRevision, err := updateAPI(vhost, apiInfo, apiProject, allEnvironments)
 		if err != nil {
-			return deployedRevisionList, err
+			return deployedRevisionList, fmt.Errorf("%v:%v with UUID \"%v\"", apiInfo.Name, apiInfo.Version, apiInfo.ID)
 		}
-        if deployedRevision != nil {
+		if deployedRevision != nil {
```
GDLMadushanka, line 285: shall we fix the formatting issue

adapter/internal/api/apis_impl.go
```diff
@@ -231,8 +232,7 @@ func verifyMandatoryFields(apiJSON config.APIJsonData) error {
 
 // ApplyAPIProjectFromAPIM accepts an apictl project (as a byte array), list of vhosts with respective environments
 // and updates the xds servers based upon the content.
-func ApplyAPIProjectFromAPIM(payload []byte, vhostToEnvsMap map[string][]string) ([]*notifier.DeployedAPIRevision,error) {
-	var deployedRevisionList []*notifier.DeployedAPIRevision
+func ApplyAPIProjectFromAPIM(payload []byte, vhostToEnvsMap map[string][]string) (deployedRevisionList []*notifier.DeployedAPIRevision, err error) {
```
GDLMadushanka, line 235: shall we fix the formatting issue (line length)

adapter/internal/api/apis_impl.go
```diff
@@ -337,12 +354,22 @@ func ApplyAPIProjectInStandaloneMode(payload []byte, override *bool) error {
 
 	// TODO: (renuka) optimize to update cache only once when all internal memory maps are updated
 	for vhost, environments := range vhostToEnvsMap {
-		updateAPI(vhost, apiInfo, apiProject, environments)
+		_, err := updateAPI(vhost, apiInfo, apiProject, environments)
+		if err != nil {
+			return err
+		}
 	}
 	return nil
```
GDLMadushanka, line 362: can we use just return here since we have named return parameters ?

adapter/internal/api/apis_impl.go
ashera96: This was a fix for a formatting issue that already existed.

adapter/internal/api/apis_impl.go
ashera96: Fixed with https://github.com/wso2/product-microgateway/pull/2482

adapter/internal/api/apis_impl.go
ashera96: Fixed with https://github.com/wso2/product-microgateway/pull/2482

Issue #2132, opened 2021-07-05
Update Local Adapter to communicate with Global adapter


Jayanie:
### Describe your problem

Local Adapter needs to be changed to communicate with Global Adapter. 

### Describe your solution

At the Local Adapter startup, API UUID list should be received from the Global Adapter.
Local Adapter needs to fetch the corresponding APIs from the control plane at startup. (using /retrieve-api-artifacts endpoint)
When a API deploy or undeploy event is received from the Local Adapter, Enforcer and Router components need to be updated.

Related Issue: https://github.com/wso2-enterprise/choreo/issues/4333

Jayanie:
This is fixed in https://github.com/wso2/product-microgateway/pull/2127

PR #2127, opened 2021-07-05
Local Adapter Related Implementation

Jayanie:
### Purpose
This PR includes:
 1. Local Adapter startup and fetching the APIs from the control plane with API UUID list
 2. Handling API events received from the GA (Deploying and removing APIs)
 
### Issues
Fixes #2132
Fixes https://github.com/wso2-enterprise/choreo/issues/4333
Main Task: https://github.com/wso2-enterprise/choreo/issues/3930

---
#### Maintainers: Check before merge
- [x] Assigned 'Type' label
- [x] Assigned the project
- [x] Validated respective github issues
- [x] Assigned milestone to the github issue(s)

PR #2148, opened 2021-07-11
Include organization UUID within the protobuf payload from global adapter 

VirajSalaka:
### Purpose
The organization ID is required for deletion of APIs as the APIs are located within a map referred by organizationID. Without this, the API UUID is required to be searched under all organization IDs prior to deletion. 

### Issues

Related to #2132

### Automation tests
 - Unit tests added: No
 - Integration tests added: No

### Tested environments

Not Tested

---
- [x] Assigned 'Type' label
- [x] Assigned the project


mpmunasinghe (approved):
LGTM

adapter/internal/ga/client.go
```diff
@@ -64,13 +64,14 @@ const (
 // APIEvent represents the event corresponding to a single API Deploy or Remove event
 // based on XDS state changes
 type APIEvent struct {
-	APIUUID       string
-	RevisionUUID  string
-	IsDeployEvent bool
+	APIUUID          string
+	RevisionUUID     string
+	IsDeployEvent    bool
+	OrganizationUUID string
```
Rajith90, line 70: These are inside the internal package, internal package is only used by LA, so what would be the org ID when used on premise

VirajSalaka: Yes. Because this event is only used by local adapter. In JMS scenario we read amqp delivery object, convert into an internal object proceed. Just like that from the XDS content, this event would be populated and pushed into a channel.

VirajSalaka: This event is used by https://github.com/wso2/product-microgateway/pull/2127

Rajith90: My concerns is we are bringing Choreo specific logic to the LA

VirajSalaka: Sorry my answer was not complete. The orgID used in on premise scenario would be,

- In api deployment flow, it is fetched from the api.json (Hence we do not need for GA-LA flow as well, but added to maintain consistency)
- In api Remove from gateway flow, this org-id is picked from `tenant-domain` in the JMS event. (This is what we are trying to get by introducing this Org-id)
https://github.com/wso2/product-microgateway/blob/a0e2d70161b345935743ea7bbac311aae296f974/adapter/internal/messaging/notification_listener.go#L172

Rajith90 (approved):

choreo-cicd:
[failed] dev-deployment : 20210825.10

choreo-cicd:
[succeeded] prod-deployment : 20210831.1

Issue #2165, opened 2021-07-20
CORS Headers are not showing


Sumudu-Sahan:
### Description:
CORS headers are not showing in the token response even after enable the CORS configurations via the **toolkit-config.toml** file.

### Steps to reproduce:

- Create an MGW project.
- Open the **toolkit-config.toml** file inside the <MGW_TOOLKIT_HOME>/conf directory and enable the CORS headers configurations suggested in the documentation [1] as below.

```
[corsConfiguration]
accessControlAllowOrigins = ["*"]
accessControlAllowCredentials = false
corsConfigurationEnabled = true
accessControlAllowHeaders = ["authorization", "Access-Control-Allow-Origin", "Content-Type", "SOAPAction"]
accessControlAllowMethods = ["GET", "PUT", "POST", "DELETE", "PATCH", "OPTIONS"]
accessControlExposeHeaders = []
```

- Build the project and deploy the runtime artifact by using MGW runtime.
-  Invoke the token endpoint by using the **OPTIONS** method by parsing **Origin**, **Access-Control-Request-Headers**, and **Access-Control-Request-Method** headers in the token request as below.
![CORS_HEADERS](https://user-images.githubusercontent.com/18566908/126327211-d9f4f500-3c97-415c-9aab-2d1e1a1a1e87.png)
- But able to see only the **access-control-allow-origin** header in the response above.

[1] https://mg.docs.wso2.com/en/latest/how-tos/security/enabling-cors/#enable-cors-configuration-for-endpoints

### Affected Product Version:
3.2.0

PR #2173, opened 2021-07-22
Fix missing cors headers in token & other services

chashikajw:
## Purpose
As $subject this will fix the missing cors headers in /token , /revoke, /userInfo, /authorize, /apiKey services

## Approach
Before the fix, it doesn't define the extract method for all the service resources.
Hence ballerina treats Option call also as a supported method in particular resources.

As a fix, we have defined the extract method when creating particular service resources.

## Fixes
https://github.com/wso2/product-microgateway/issues/2165


Rajith90 (approved):

chashikajw:
This was sent to 3.2.x branch,

PR #2177, opened 2021-07-27
Initial skeleton to test azure event subscription flow

dnwick:
### Purpose

This is the initial code segment which reads environment feature flag ('FEAURE_FLAG_REPLACE_EVENT_HUB') and starts going through that flow. This PR is to test that initial flow. See sub task https://github.com/wso2/product-microgateway/issues/2186

Fixes: https://github.com/wso2/product-microgateway/issues/2186

### Automation tests
 - Unit tests added: No
 - Integration tests added: No

### Tested environments

Not Tested

---
- [x] Assigned 'Type' label
- [x] Assigned the project

### DIFF
--- a/adapter/internal/adapter/adapter.go
+++ b/adapter/internal/adapter/adapter.go
@@ -231,17 +231,17 @@ func Run(conf *config.Config) {
 		if (envValue != "") {
 			isAzureEventingFeatureFlagEnabled, err = strconv.ParseBool(envValue)
 			if (err != nil) {
-				logger.LoggerMgw.Error("Error occurred while parsing FEAURE_FLAG_REPLACE_EVENT_HUB environment value.",
-					err)
+				logger.LoggerMgw.Info("[TEST][FEATURE_FLAG_REPLACE_EVENT_HUB] Error occurred while parsing " +
+					"FEAURE_FLAG_REPLACE_EVENT_HUB environment value.", err)
 			}
 		}
 
 		if (isAzureEventingFeatureFlagEnabled) {
 			logger.LoggerMgw.Info("[TEST][FEATURE_FLAG_REPLACE_EVENT_HUB] Starting to integrate with azure service bus")
 			go messaging.InitiateAndProcessEvents(conf)
-		} else {
-			go messaging.ProcessEvents(conf)
 		}
+
+		go messaging.ProcessEvents(conf)
 
 		// Fetch APIs from control plane
 		fetchAPIsOnStartUp(conf)
--- a/adapter/internal/adapter/adapter.go
+++ b/adapter/internal/adapter/adapter.go
@@ -232,7 +232,7 @@ func Run(conf *config.Config) {
 			isAzureEventingFeatureFlagEnabled, err = strconv.ParseBool(envValue)
 			if (err != nil) {
 				logger.LoggerMgw.Info("[TEST][FEATURE_FLAG_REPLACE_EVENT_HUB] Error occurred while parsing " +
-					"FEAURE_FLAG_REPLACE_EVENT_HUB environment value.", err)
+					"FEATURE_FLAG_REPLACE_EVENT_HUB environment value.", err)
 			}
 		}
 
--- a/adapter/internal/adapter/adapter.go
+++ b/adapter/internal/adapter/adapter.go
@@ -237,6 +237,7 @@ func Run(conf *config.Config) {
 		}
 
 		if (isAzureEventingFeatureFlagEnabled) {
+			logger.LoggerMgw.Info("[TEST][FEATURE_FLAG_REPLACE_EVENT_HUB] Starting to integrate with azure service bus")
 			go messaging.InitiateAndProcessEvents(conf)
 		} else {
 			go messaging.ProcessEvents(conf)
--- a/adapter/internal/adapter/adapter.go
+++ b/adapter/internal/adapter/adapter.go
@@ -227,7 +227,7 @@ func Run(conf *config.Config) {
 		var isAzureEventingFeatureFlagEnabled bool
 		var err error
 
-		envValue := os.Getenv("FEAURE_FLAG_REPLACE_EVENT_HUB")
+		envValue := os.Getenv("FEATURE_FLAG_REPLACE_EVENT_HUB")
 		if (envValue != "") {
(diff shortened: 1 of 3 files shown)


adapter/internal/adapter/adapter.go
```diff
@@ -223,8 +224,23 @@ func Run(conf *config.Config) {
...
 		// Load subscription data
 		eventhub.LoadSubscriptionData(conf)
+		var isAzureEventingFeatureFlagEnabled bool
+		var err error
+
+		envValue := os.Getenv("FEAURE_FLAG_REPLACE_EVENT_HUB")
+		if (envValue != "") {
+			isAzureEventingFeatureFlagEnabled, err = strconv.ParseBool(envValue)
+			if (err != nil) {
+				logger.LoggerMgw.Error("Error occurred while parsing FEAURE_FLAG_REPLACE_EVENT_HUB environment value.",
```
sdkottegoda, line 234: shall we make this an info log in the form we discussed? Because we don't need to hi any alerting systems as this is a test feature.

adapter/internal/adapter/adapter.go
```diff
@@ -223,8 +224,24 @@ func Run(conf *config.Config) {
...
+					err)
+			}
+		}
 
-		go messaging.ProcessEvents(conf)
+		if (isAzureEventingFeatureFlagEnabled) {
+			logger.LoggerMgw.Info("[TEST][FEATURE_FLAG_REPLACE_EVENT_HUB] Starting to integrate with azure service bus")
+			go messaging.InitiateAndProcessEvents(conf)
+		} else {
```
sdkottegoda, line 242: This should not be an else, it needs to be done anyway. Because we don't need out feature to break the existing functionality.

adapter/internal/adapter/adapter.go
dnwick: sure

adapter/pkg/messaging/azure_connection.go
```diff
@@ -0,0 +1,46 @@
...
+// Package messaging holds the implementation for event listeners functions
+package messaging
+
+import (
+	logger "github.com/wso2/product-microgateway/adapter/internal/loggers"
+)
+
+var (
+	// AzureRevokedTokenChannel stores the revoked token events
+	AzureRevokedTokenChannel chan []byte
```
sdkottegoda, line 27: Should we rename this to channel?

adapter/pkg/messaging/azure_connection.go
```diff
@@ -0,0 +1,46 @@
...
+)
+
+func init() {
+	AzureRevokedTokenChannel = make(chan []byte)
+}
+
+// connectToAzureServiceBus function tries to connect to the azure service bus server as long as it takes to
+// establish a connection
+func connectToAzureServiceBus() {
```
sdkottegoda, line 36: Do we need 2 methods here?

malinthaprasan (approved):
Please get @Rajith90's approval for merging.

adapter/pkg/messaging/azure_connection.go
pubudu538: +1.. We should use generic terms.

dnwick: We need channels for each topic like tokenRevocation, Notification etc and generic term would be for token revocation would be RevokedTokenChannel but this is already used in earlier implementation. Since azure_connection.go is in same package we cannot use that. wdyt ?

adapter/internal/adapter/adapter.go
dnwick: fixed

adapter/pkg/messaging/azure_connection.go
sdkottegoda: How about asbRevokedTokenChannel? Because, using 'chan' here doesn't anyway resolve the problem. You'll face the same issue with the next topic.

dnwick: chan is a keyword and this how channels are getting created in go. eg- channel1 chan string, channel2 chan string  are two channels of string type.
Concern is the name AzureRevokedTokenChannel i guess

adapter/internal/adapter/adapter.go
```diff
@@ -223,6 +224,22 @@ func Run(conf *config.Config) {
 	if eventHubEnabled {
 		// Load subscription data
 		eventhub.LoadSubscriptionData(conf)
+		var isAzureEventingFeatureFlagEnabled bool
+		var err error
+
+		envValue := os.Getenv("FEATURE_FLAG_REPLACE_EVENT_HUB")
```
Rajith90, line 230: Shall we use a constant for this

adapter/internal/adapter/adapter.go
```diff
@@ -223,6 +224,22 @@ func Run(conf *config.Config) {
...
+		if (envValue != "") {
+			isAzureEventingFeatureFlagEnabled, err = strconv.ParseBool(envValue)
+			if (err != nil) {
+				logger.LoggerMgw.Info("[TEST][FEATURE_FLAG_REPLACE_EVENT_HUB] Error occurred while parsing " +
```
Rajith90, line 234: Even this is a temp code, lets use the logger.Error

adapter/internal/adapter/adapter.go
```diff
@@ -223,6 +224,22 @@ func Run(conf *config.Config) {
 	if eventHubEnabled {
 		// Load subscription data
 		eventhub.LoadSubscriptionData(conf)
+		var isAzureEventingFeatureFlagEnabled bool
+		var err error
+
+		envValue := os.Getenv("FEATURE_FLAG_REPLACE_EVENT_HUB")
+		if (envValue != "") {
+			isAzureEventingFeatureFlagEnabled, err = strconv.ParseBool(envValue)
```
Rajith90, line 232: Please add TODOs with your github  handle(user name) to remove these flag related code when feature is completed.

Rajith90 (approved):

choreo-cicd:
[succeeded] dev-deployment : 20210802.1

choreo-cicd:
[succeeded] stage-deployment : 20210806.1

choreo-cicd:
[succeeded] prod-deployment : 20210809.1

Issue #2185, opened 2021-07-30
Implement AzureEventListener in local adaptor to receive messages from Azure Service Bus for token revocation and notification events


dnwick:
### Describe your problem(s)
Implement AzureEventListener in local adaptor to receive messages from Azure Service Bus for token revocation and notification events

Use acknowledgement for reliable message delivery
Verify connection retries

### How will you implement it

---

PR #2194, opened 2021-08-09
Add subscriber implementation for token revocation and notification events from azure service bus

dnwick:
### Purpose
Add implementation for receiving events for token revocation and notification topics from azure service bus

### Issues
Fixes https://github.com/wso2/product-microgateway/issues/2185

### Automation tests
 - Unit tests added: Yes
 - Integration tests added: No

### Tested environments

locally tested

---
- [x] Assigned 'Type' label
- [x] Assigned the project


adapter/internal/adapter/adapter.go
```diff
@@ -230,15 +230,15 @@ func Run(conf *config.Config) {
 
 		// TODO: (dnwick) remove env variable once the feature is complete
 		envValue := os.Getenv(featureFlagReplaceEventHub)
```
Rajith90, line 232: Lets use proper name, to give the idea env Value for what

adapter/internal/messaging/notification_listener.go
```diff
@@ -105,6 +105,21 @@ func handleNotification() {
...
 }
 
+func handleAzureNotification() {
+	for d := range msg.AzureNotificationChannel {
+		logger.LoggerInternalMsg.Infof("[TEST][FEATURE_FLAG_REPLACE_EVENT_HUB] message received for " +
+			"NotificationChannel = " + string(d))
+		var notification msg.EventNotification
+		error := parseNotificationJSONEvent(d, &notification)
+		if error != nil {
+			continue
```
Rajith90, line 115: Lets handle the errors properly, retrying and etc

adapter/pkg/messaging/azure_connection.go
```diff
@@ -20,22 +20,64 @@ package messaging
...
+		topicManager := namespace.NewTopicManager()
+		ctx, cancel := context.WithCancel(context.Background())
+		defer cancel()
+		availableTopics, err = topicManager.List(ctx)
+		if err != nil {
+			logger.LoggerMgw.Errorf("[TEST][FEATURE_FLAG_REPLACE_EVENT_HUB] Error occurred while trying to get topic " +
+				"list from azure service bus :%v", err)
+		}
+		logger.LoggerMgw.Info("[TEST][FEATURE_FLAG_REPLACE_EVENT_HUB] Successfully received topic list ")
+	} else {
```
Rajith90, line 65: We should retry these cases

adapter/pkg/messaging/azure_connection.go
```diff
@@ -20,22 +20,64 @@ package messaging
...
-	return err
+	namespace, err := servicebus.NewNamespace(servicebus.NamespaceWithConnectionString(eventListeningEndpoint))
+
+	if err == nil {
+		logger.LoggerMgw.Info("[TEST][FEATURE_FLAG_REPLACE_EVENT_HUB] Successfully received namespace ")
+		topicManager := namespace.NewTopicManager()
+		ctx, cancel := context.WithCancel(context.Background())
+		defer cancel()
+		availableTopics, err = topicManager.List(ctx)
+		if err != nil {
```
Rajith90, line 60: We should retry these cases

adapter/pkg/messaging/azure_listener.go
```diff
@@ -0,0 +1,160 @@
...
+
+	var topicExistForFurtherProcess bool
+	var subscriptionExistForFurtherProcess bool
+	parentContext := context.Background()
+	logger.LoggerMgw.Info("[TEST][FEATURE_FLAG_REPLACE_EVENT_HUB] Starting broker consumer for topic name : " +
+		topicName)
+
+	//In Azure service bus subscription names can contain letters, numbers, periods (.), hyphens (-), and
+	// underscores (_), up to 50 characters. Subscription names are also case-insensitive.
+	var subscriptionName = topicName + "_" + componentName + "_sub"
```
Rajith90, line 40: When multiple adapters starting up, using the same subscription name would cause issues right?

adapter/internal/messaging/notification_listener.go
dnwick: if there is an error here means there is an issue with the consumed event. So we should just log and continue right ? Retry should happen receiving end before message is given to the channel ?

adapter/pkg/messaging/azure_connection.go
dnwick: Retry will be added next .. will add todo's

adapter/pkg/messaging/azure_listener.go
dnwick: When multiple adapters are there then for each messages will be delivered in round robin manner

adapter/pkg/messaging/azure_listener.go
```diff
@@ -0,0 +1,160 @@
...
+
+				logger.LoggerMgw.Info("[TEST][FEATURE_FLAG_REPLACE_EVENT_HUB] Starting to receive messages for " +
+					"subscriptionName  " + subscriptionName + " from azure service bus for topic name " + topicName)
+
+				ctx, cancel := context.WithCancel(parentContext)
+				defer cancel()
+				err = topicSubscriptionClient.Receive(ctx, servicebus.HandlerFunc(func(ctx context.Context,
+					message *servicebus.Message) error {
+					dataChannel <- message.Data
+					<-ackChannel
```
VirajSalaka, line 128: The process will halt if an error occurs in channel consuming part it seems. As the there is a continue statement within loop prior to sending an Ack through the channel.

dnwick: Ya true this is just to test till event consumption. Basically we need to <-ackChannel before doing continue also . I kept it till i finish other logic but will add a todo

adapter/pkg/messaging/azure_listener.go
dnwick: As per Viraj do we need a broadcast behaviour here ?

dnwick: We can have a uuid or ask the deployment to give a id as env variable and create a new subscription so that each adapther will get same msg

adapter/internal/messaging/azure_listener.go
```diff
@@ -22,13 +22,30 @@ import (
...
-	msg.InitiateBrokerConnection(config.ControlPlane.ASBConnectionParameters.EventListeningEndpoint)
-
+	namespace, availableTopicList, err := msg.InitiateBrokerConnectionAndGetAvailableTopics(
+		config.ControlPlane.ASBConnectionParameters.EventListeningEndpoint)
+	health.SetControlPlaneBrokerStatus(err == nil)
+	if err == nil {
+		logger.LoggerMgw.Info("[TEST][FEATURE_FLAG_REPLACE_EVENT_HUB] Initiated broker connection successfully ")
+		msg.InitiateConsumers(namespace, availableTopicList, componentName, subscriptionIdleTimeDuration)
+	}
+	go handleAzureNotification()
```
sdkottegoda, line 49: I suppose we shouldn't continue if there's any error?

dnwick: this is done

sdkottegoda (approved):

Rajith90 (approved):

adapter/pkg/messaging/azure_connection.go
dnwick: as mentioned retry will be added next phase

sdkottegoda:
Manually verified the dev deployment and it is successful for choreo-connect-adapter

choreo-cicd:
[succeeded] stage-deployment : 20210811.3

choreo-cicd:
[succeeded] prod-deployment : 20210812.1

Issue #2195, opened 2021-08-11
Incorporate retry logic for azure subscriber logic 


dnwick:
### Describe your problem(s)
Retry mechanism should be applied in both connection initialization and message receive flow

### How will you implement it

---

PR #2196, opened 2021-08-11
Add connection retry logic for service bus subscription

dnwick:
### Purpose
This PR will improve the subscription logic to handle any connection errors and retry according to the policy

### Issues
Fixes https://github.com/wso2/product-microgateway/issues/2195

### Automation tests
 - Unit tests added: No
 - Integration tests added: No

### Tested environments

Not Tested

---
- [x] Assigned 'Type' label
- [x] Assigned the project

### DIFF
--- a/adapter/config/default_config.go
+++ b/adapter/config/default_config.go
@@ -252,7 +252,7 @@ var defaultConfig = &Config{
 		},
 		ASBConnectionParameters: asbConnectionParameters{
 			EventListeningEndpoint: "$env{ASB_CONNECTION_STRING}",
-			ReconnectInterval:      5000,
+			ReconnectInterval:      5000, //in milli seconds
 			ReconnectRetryCount:    60,
 		},
 	},
--- a/adapter/config/default_config.go
+++ b/adapter/config/default_config.go
@@ -252,8 +252,8 @@ var defaultConfig = &Config{
 		},
 		ASBConnectionParameters: asbConnectionParameters{
 			EventListeningEndpoint: "$env{ASB_CONNECTION_STRING}",
-			ReconnectInterval:      30,
-			ReconnectRetryCount:    10,
+			ReconnectInterval:      5000,
+			ReconnectRetryCount:    60,
 		},
 	},
 	Analytics: analytics{
--- a/adapter/config/default_config.go
+++ b/adapter/config/default_config.go
@@ -251,7 +251,10 @@ var defaultConfig = &Config{
 			EventListeningEndpoints: []string{"amqp://admin:$env{cp_admin_pwd}@apim:5672?retries='10'&connectdelay='30'"},
 		},
 		ASBConnectionParameters: asbConnectionParameters{
-			EventListeningEndpoint: "$env{ASB_CONNECTION_STRING}"},
+			EventListeningEndpoint: "$env{ASB_CONNECTION_STRING}",
+			ReconnectInterval:      30,
+			ReconnectRetryCount:    10,
+		},
 	},
 	Analytics: analytics{
 		Enabled: false,
--- a/adapter/internal/messaging/azure_listener.go
+++ b/adapter/internal/messaging/azure_listener.go
@@ -20,7 +20,6 @@ package messaging
 
 import (
 	"time"
-
 	servicebus "github.com/Azure/azure-service-bus-go"
 	"github.com/wso2/product-microgateway/adapter/config"
 	logger "github.com/wso2/product-microgateway/adapter/internal/loggers"
@@ -29,8 +28,11 @@ import (
 )
 
 const (
-	componentName                string = "adapter"
-	subscriptionIdleTimeDuration        = time.Duration(72 * time.Hour)
+	componentName                              = "adapter"
+	subscriptionIdleTimeDuration               = time.Duration(72 * time.Hour)
+	defaultReconnectRetryCount                 = 10
+	defaultReconnectInterval     time.Duration = 10
+
 )
 
 // InitiateAndProcessEvents to pass event consumption
@@ -43,13 +45,14 @@ func InitiateAndProcessEvents(config *config.Config) {
 	logger.LoggerMgw.Info("[TEST][FEATURE_FLAG_REPLACE_EVENT_HUB] EventListeningEndpoint is ",
 		config.ControlPlane.ASBConnectionParameters.EventListeningEndpoint)
 	if reconnectRetryCount == 0 {
-		reconnectRetryCount = 10
+		reconnectRetryCount = defaultReconnectRetryCount
 	}
 	if reconnectInterval == 0 {
-		reconnectInterval = 30
+		reconnectInterval = defaultReconnectInterval
 	}
 	namespace, availableTopicList, err := msg.InitiateBrokerConnectionAndGetAvailableTopics(
-		config.ControlPlane.ASBConnectionParameters.EventListeningEndpoint, reconnectRetryCount, reconnectInterval*time.Second)
+		config.ControlPlane.ASBConnectionParameters.EventListeningEndpoint, reconnectRetryCount,
+		reconnectInterval * time.Second)
 	health.SetControlPlaneBrokerStatus(err == nil)
 	if err == nil {
 		logger.LoggerMgw.Info("[TEST][FEATURE_FLAG_REPLACE_EVENT_HUB] Initiated broker connection successfully ")
--- a/adapter/internal/messaging/azure_listener.go
+++ b/adapter/internal/messaging/azure_listener.go
@@ -20,7 +20,6 @@ package messaging
 
 import (
 	"time"
-	servicebus "github.com/Azure/azure-service-bus-go"
 	"github.com/wso2/product-microgateway/adapter/config"
 	logger "github.com/wso2/product-microgateway/adapter/internal/loggers"
 	"github.com/wso2/product-microgateway/adapter/pkg/health"
@@ -38,7 +37,6 @@ const (
 // InitiateAndProcessEvents to pass event consumption
 func InitiateAndProcessEvents(config *config.Config) {
 	var err error
-	var namespace *servicebus.Namespace
 	var reconnectRetryCount = config.ControlPlane.ASBConnectionParameters.ReconnectRetryCount
 	var reconnectInterval = config.ControlPlane.ASBConnectionParameters.ReconnectInterval
 	logger.LoggerMgw.Info("[TEST][FEATURE_FLAG_REPLACE_EVENT_HUB] Starting InitiateAndProcessEvents method")
@@ -50,14 +48,14 @@ func InitiateAndProcessEvents(config *config.Config) {
 	if reconnectInterval == 0 {
 		reconnectInterval = defaultReconnectInterval
 	}
-	namespace, availableTopicList, err := msg.InitiateBrokerConnectionAndGetAvailableTopics(
-		config.ControlPlane.ASBConnectionParameters.EventListeningEndpoint, reconnectRetryCount,
-		reconnectInterval * time.Second)
+	subscriptionMetaDataList, err := msg.InitiateBrokerConnectionAndValidate(
+		config.ControlPlane.ASBConnectionParameters.EventListeningEndpoint, componentName, reconnectRetryCount,
+		reconnectInterval * time.Second, subscriptionIdleTimeDuration)
 	health.SetControlPlaneBrokerStatus(err == nil)
 	if err == nil {
-		logger.LoggerMgw.Info("[TEST][FEATURE_FLAG_REPLACE_EVENT_HUB] Initiated broker connection successfully ")
-		msg.InitiateConsumers(namespace, availableTopicList, componentName, subscriptionIdleTimeDuration,
-			reconnectRetryCount, reconnectInterval*time.Second)
+		logger.LoggerMgw.Info("[TEST][FEATURE_FLAG_REPLACE_EVENT_HUB] Initiated broker connection and meta " +
+			"data creation successfully ")
+		msg.InitiateConsumers(subscriptionMetaDataList)
 		go handleAzureNotification()
 		go handleAzureTokenRevocation()
 	}
--- a/adapter/internal/messaging/azure_listener.go
+++ b/adapter/internal/messaging/azure_listener.go
@@ -19,32 +19,42 @@
 package messaging
 
 import (
+	"time"
+
+	servicebus "github.com/Azure/azure-service-bus-go"
 	"github.com/wso2/product-microgateway/adapter/config"
 	logger "github.com/wso2/product-microgateway/adapter/internal/loggers"
-	msg "github.com/wso2/product-microgateway/adapter/pkg/messaging"
-	servicebus "github.com/Azure/azure-service-bus-go"
 	"github.com/wso2/product-microgateway/adapter/pkg/health"
-	"time"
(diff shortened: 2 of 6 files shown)


adapter/config/default_config.go
```diff
@@ -251,7 +251,10 @@ var defaultConfig = &Config{
 			EventListeningEndpoints: []string{"amqp://admin:$env{cp_admin_pwd}@apim:5672?retries='10'&connectdelay='30'"},
 		},
 		ASBConnectionParameters: asbConnectionParameters{
-			EventListeningEndpoint: "$env{ASB_CONNECTION_STRING}"},
+			EventListeningEndpoint: "$env{ASB_CONNECTION_STRING}",
+			ReconnectInterval:      30,
+			ReconnectRetryCount:    10,
```
sdkottegoda, line 256: I think the default should be -1 indicating infinite

adapter/pkg/messaging/azure_connection.go
```diff
@@ -37,45 +48,175 @@ func init() {
...
 		topicManager := namespace.NewTopicManager()
-		ctx, cancel := context.WithCancel(context.Background())
-		defer cancel()
-		availableTopics, err = topicManager.List(ctx)
-		//TODO (dnwick) (add retry logic)
-		if err != nil {
-			logger.LoggerMgw.Errorf("[TEST][FEATURE_FLAG_REPLACE_EVENT_HUB] Error occurred while trying to get topic " +
-				"list from azure service bus :%v", err)
+		var availableTopics []*servicebus.TopicEntity
+		for j := 0; j < reconnectRetryCount; j++ {
```
sdkottegoda, line 65: or if reconnectRetryCount == -1?

adapter/pkg/messaging/azure_connection.go
```diff
@@ -37,45 +48,175 @@ func init() {
...
+			ctx, cancel := context.WithCancel(context.Background())
+			defer cancel()
+			availableTopics, getTopicListError = topicManager.List(ctx)
+			if getTopicListError != nil {
+				logger.LoggerMgw.Errorf("[TEST][FEATURE_FLAG_REPLACE_EVENT_HUB] Error occurred while trying to get topic "+
+					"list from azure service bus :%v. Retrying after %d seconds", err, reconnectInterval)
```
sdkottegoda, line 71: Let's log the retry attempt if it's greater than 0

adapter/config/default_config.go
sdkottegoda: Not needed, the current default is fine

resources/conf/config.toml.template
```diff
@@ -366,6 +366,8 @@ idleTimeoutInSeconds = 3600
   # TODO: (dnwick) remove this config once a common connection parameter is introduced
   [controlPlane.asbConnectionParameters]
     eventListeningEndpoint = "$env{ASB_CONNECTION_STRING}"
+    reconnectInterval = 30
```
sdkottegoda, line 369: Need to change the template?

adapter/internal/messaging/azure_listener.go
```diff
@@ -19,32 +19,42 @@
...
 	var err error
-	var namespace *servicebus.Namespace
+	var reconnectRetryCount = config.ControlPlane.ASBConnectionParameters.ReconnectRetryCount
+	var reconnectInterval = config.ControlPlane.ASBConnectionParameters.ReconnectInterval
 	logger.LoggerMgw.Info("[TEST][FEATURE_FLAG_REPLACE_EVENT_HUB] Starting InitiateAndProcessEvents method")
 	logger.LoggerMgw.Info("[TEST][FEATURE_FLAG_REPLACE_EVENT_HUB] EventListeningEndpoint is ",
 		config.ControlPlane.ASBConnectionParameters.EventListeningEndpoint)
-	namespace, availableTopicList, err := msg.InitiateBrokerConnectionAndGetAvailableTopics(
-		config.ControlPlane.ASBConnectionParameters.EventListeningEndpoint)
+	if reconnectRetryCount == 0 {
```
sdkottegoda, line 44: Don't we get nil if undefined? My concern is, this makes 0 an invalid argument

adapter/pkg/messaging/azure_connection.go
```diff
@@ -37,45 +50,180 @@ func init() {
...
+		go func(subscriptionMetaData Subscription) {
+			logger.LoggerMgw.Info("[TEST][FEATURE_FLAG_REPLACE_EVENT_HUB] starting the consumer for topic : " +
+				subscriptionMetaData.topicName)
+			startBrokerConsumer(subscriptionMetaData, reconnectInterval)
+			select {}
+		}(subscriptionMetaData)
+	}
+}
 
+func validateAndCreateTopicForSubscription(availableTopicList []*servicebus.TopicEntity, ns *servicebus.Namespace) error {
```
sdkottegoda, line 144: can we simply say createTopicIfNotExists

adapter/pkg/messaging/azure_connection.go
```diff
@@ -37,45 +50,180 @@ func init() {
...
+			if processError != nil {
+				retryAttemptMessage = ""
+				if reconnectRetryCount > 0 {
+					retryAttemptMessage = "Retry attempt : " + strconv.Itoa(reconnectRetryCount)
+				}
+				logger.LoggerMgw.Errorf("[TEST][FEATURE_FLAG_REPLACE_EVENT_HUB] %v " + retryAttemptMessage +
+					" Retrying after %s", processError, reconnectInterval)
+				time.Sleep(reconnectInterval)
+				continue
+			}
```
sdkottegoda, line 101: IIUC, this block should be identical to line 75 -> 87 with the operation changed. Shall we extract to a common method?

sdkottegoda: We might need to change the method definition to return status, error

adapter/pkg/messaging/azure_connection.go
```diff
@@ -37,45 +50,180 @@ func init() {
...
+			logger.LoggerMgw.Info("[TEST][FEATURE_FLAG_REPLACE_EVENT_HUB] Topic " + key + " created")
+		} else {
+			logger.LoggerMgw.Info("[TEST][FEATURE_FLAG_REPLACE_EVENT_HUB] Topic " + key + " Exist ")
+		}
+	}
+	return nil
+}
+
+func validateAndGetSubscriptionMetaDataList(metaDataList []Subscription, ns *servicebus.Namespace, reconnectRetryCount int,
```
sdkottegoda, line 171: retrieveSubscriptionMetadata?

adapter/pkg/messaging/azure_connection.go
dnwick: done

adapter/pkg/messaging/azure_connection.go
dnwick: done

adapter/pkg/messaging/azure_connection.go
```diff
@@ -37,45 +50,180 @@ func init() {
...
+			if processError != nil {
+				retryAttemptMessage = ""
+				if reconnectRetryCount > 0 {
+					retryAttemptMessage = "Retry attempt : " + strconv.Itoa(reconnectRetryCount)
+				}
+				logger.LoggerMgw.Errorf("[TEST][FEATURE_FLAG_REPLACE_EVENT_HUB] %v. " + retryAttemptMessage + " " +
+					" Retrying after %s ", processError, reconnectInterval)
+				time.Sleep(reconnectInterval)
+				continue
+			}
```
sdkottegoda, line 115: same as above

adapter/pkg/messaging/azure_connection.go
dnwick: +1

adapter/pkg/messaging/azure_listener.go
```diff
@@ -19,126 +19,49 @@
...
 	}
+	parentContext := context.Background()
 
-	if topicExistForFurtherProcess {
-		logger.LoggerMgw.Info("[TEST][FEATURE_FLAG_REPLACE_EVENT_HUB] Topic " + topicName + " exist. Can proceed")
-		subManager, err := ns.NewSubscriptionManager(topicName)
-
+	for {
+		//topic subscription client creation
+		topicSubscriptionClient, err := subscriptionMetaData.subscriptionManager.Topic.NewSubscription(subscriptionName)
```
sdkottegoda, line 43: If something goes wrong in the connection, will the subscriptionManager still be valid? Or should we reset that?

dnwick: it will be valid .. I tested it

adapter/pkg/messaging/azure_connection.go
dnwick: +1

resources/conf/config.toml.template
dnwick: Done

adapter/pkg/messaging/azure_connection.go
dnwick: Done

adapter/pkg/messaging/azure_connection.go
dnwick: Done

adapter/internal/messaging/azure_listener.go
dnwick: Will check and update

sdkottegoda (approved):

adapter/internal/messaging/azure_listener.go
```diff
@@ -19,32 +19,42 @@
...
+	msg "github.com/wso2/product-microgateway/adapter/pkg/messaging"
 )
 
 const (
-	componentName string = "adapter"
-	subscriptionIdleTimeDuration = time.Duration(72 * time.Hour)
+	componentName                              = "adapter"
+	subscriptionIdleTimeDuration               = time.Duration(72 * time.Hour)
+	defaultReconnectRetryCount                 = 60
+	defaultReconnectInterval     time.Duration = 5000
```
Rajith90, line 33: Shall we add a comment that this in ms

adapter/internal/messaging/azure_listener.go
```diff
@@ -19,32 +19,42 @@
...
 	logger.LoggerMgw.Info("[TEST][FEATURE_FLAG_REPLACE_EVENT_HUB] Starting InitiateAndProcessEvents method")
 	logger.LoggerMgw.Info("[TEST][FEATURE_FLAG_REPLACE_EVENT_HUB] EventListeningEndpoint is ",
 		config.ControlPlane.ASBConnectionParameters.EventListeningEndpoint)
-	namespace, availableTopicList, err := msg.InitiateBrokerConnectionAndGetAvailableTopics(
-		config.ControlPlane.ASBConnectionParameters.EventListeningEndpoint)
+	if reconnectRetryCount == 0 {
+		reconnectRetryCount = defaultReconnectRetryCount
+	}
+	if reconnectInterval == 0 {
+		reconnectInterval = defaultReconnectInterval
```
Rajith90, line 48: In other places this not how we set default values. The struct is set the default values first and then we override from the config. So you don't have to check for if it is equal to zero

adapter/pkg/messaging/azure_connection.go
```diff
@@ -37,45 +50,181 @@ func init() {
...
+		}
+	}
+	logger.LoggerMgw.Info("[TEST][FEATURE_FLAG_REPLACE_EVENT_HUB] Topic " + topicName + " does not Exist ")
+	return false
+}
+
+func logError(reconnectRetryCount int, reconnectInterval time.Duration, errVal error) {
+	retryAttemptMessage := ""
+	if reconnectRetryCount > 0 {
+		retryAttemptMessage = "Retry attempt : " + strconv.Itoa(reconnectRetryCount)
```
Rajith90, line 226: Shall we include details like retry attempt to which endpoint

adapter/pkg/messaging/azure_listener.go
```diff
@@ -19,126 +19,49 @@
...
-		ctx, cancel := context.WithCancel(parentContext)
-		defer cancel()
-		_, err = subManager.Put(ctx, subscriptionName, opts...)
-
+		logger.LoggerMgw.Info("[TEST][FEATURE_FLAG_REPLACE_EVENT_HUB] Starting to receive messages for " +
+			"subscriptionName  " + subscriptionName + " from azure service bus for topic name " + topicName)
+		func() {
+			ctx, cancel := context.WithCancel(parentContext)
+			defer cancel()
+			err = topicSubscriptionClient.Receive(ctx, servicebus.HandlerFunc(func(ctx context.Context,
```
VirajSalaka, line 54: Since method has the parent context and the code is inline,
I guess we could avoid func annotation

dnwick: no this is in a for loop so since we use defer cancel() it will cancel when the for loop finishers which is never as we are retrying infinitely. So having anonymous func will make sure context is within that scope and when it goes out of it because of an error the context will also canceled else there will be a memory leak

adapter/pkg/messaging/azure_connection.go
dnwick: This method gets the errVal which will have all the information regarding the error, what is the issue and which method caused it. There are multiple processes using the logError method and I wanted to keep it generic as errVal has the all the information of the issue . Does it make sense ? Shall we keep this as it is

adapter/internal/messaging/azure_listener.go
dnwick: ok this was done in earlier broker impl. Will remove it

adapter/internal/messaging/azure_listener.go
dnwick: According to below comment we can remove this now. Will add a comment in the place we add default values

adapter/internal/messaging/azure_listener.go
dnwick: This is remove now as per below comments

adapter/internal/messaging/azure_listener.go
dnwick: fixed in https://github.com/wso2/product-microgateway/pull/2196/commits/f6486f812ca86d1d6756e920332a21758e1e99c6

adapter/internal/messaging/azure_listener.go
dnwick: fixed in https://github.com/wso2/product-microgateway/pull/2196/commits/f6486f812ca86d1d6756e920332a21758e1e99c6

adapter/pkg/messaging/azure_connection.go
```diff
@@ -37,45 +50,181 @@ func init() {
...
+				if processError != nil {
+					logError(reconnectRetryCount, reconnectInterval, processError)
+					time.Sleep(reconnectInterval)
+					continue
+				}
+			}
+			if !isTopicsCreated {
+				processError = createTopicsIfNotExist(availableTopics, namespace)
+			}
+			if processError != nil {
```
VirajSalaka, line 80: It seems this condition is never going to be reached?

VirajSalaka (approved):

Rajith90 (approved):

choreo-cicd:
[failed] dev-deployment : 20210823.15

dnwick:
Dev deployment failed due a issue in another component. This will be deployed once it is fixed

choreo-cicd:
[succeeded] stage-deployment : 20210824.5

Issue #2217, opened 2021-09-10
Inject consumed events from azure service bus to correct pipeline


dnwick:
### Describe your problem(s)
Events consumed from azure service bus need to be correctly passed to pipeline to processed correctly

### How will you implement it

---

PR #2218, opened 2021-09-10
Add logic to inject data to correct pipeline once consumed from azure service bus

dnwick:
### Purpose
Inject data to correct pipeline and config changes

### Issues
Fixes https://github.com/wso2/product-microgateway/issues/2217

### Automation tests
 - Unit tests added: No
 - Integration tests added: No

### Tested environments

Not Tested

---
- [x] Assigned 'Type' label
- [x] Assigned the project

### DIFF
--- a/adapter/internal/adapter/adapter.go
+++ b/adapter/internal/adapter/adapter.go
@@ -20,7 +20,6 @@ package adapter
 
 import (
 	"crypto/tls"
-	"strconv"
 	"strings"
 
 	discoveryv3 "github.com/envoyproxy/go-control-plane/envoy/service/discovery/v3"
@@ -74,8 +73,8 @@ var (
 )
 
 const (
-	ads                        = "ads"
-	featureFlagReplaceEventHub = "FEATURE_FLAG_REPLACE_EVENT_HUB"
+	ads          = "ads"
+	amqpProtocol = "amqp"
 )
 
 func init() {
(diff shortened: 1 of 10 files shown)


adapter/internal/adapter/adapter.go
```diff
@@ -255,11 +257,16 @@ func Run(conf *config.Config) {
 
 		if isAzureEventingFeatureFlagEnabled {
 			logger.LoggerMgw.Info("[TEST][FEATURE_FLAG_REPLACE_EVENT_HUB] Starting to integrate with azure service bus")
-			messaging.InitiateAndProcessEvents(conf)
+			var connectionURLList = conf.ControlPlane.BrokerConnectionParameters.EventListeningEndpoints
+			if strings.Contains(connectionURLList[0], amqpProtocol) {
+				go messaging.ProcessEvents(conf)
+			} else {
+				messaging.InitiateAndProcessEvents(conf)
```
Rajith90, line 264: For the eventhub its done in go routine, for ASB its not, don't we have to be consistant

adapter/internal/messaging/revoked_token_listener.go
```diff
@@ -31,48 +31,47 @@ import (
...
+		logger.LoggerInternalMsg.Infof("Event %s is received", notification.Event.PayloadData.Type)
+		logger.LoggerInternalMsg.Printf("RevokedToken: %s, Token Type: %s", notification.Event.PayloadData.RevokedToken,
 			notification.Event.PayloadData.Type)
+		processTokenRevocationEvent(&notification)
 	}
 }
 
+func processTokenRevocationEvent(notification *msg.EventTokenRevocationNotification)  {
+	var stokens []types.Resource
+	t := &keymgt.RevokedToken{}
```
Rajith90, line 63: shall we  use better variable names for t and stokens

adapter/internal/adapter/adapter.go
dnwick: InitiateAndProcessEvents will initiate connections and meta data and if any error it will avoid adapter from starting , inside InitiateAndProcessEvents after initiation is done other stuff are running relevant go routine. I do not think we need to run this also as a go routine

dnwick: also ProcessEvents will anyway going to be depricated

adapter/internal/messaging/revoked_token_listener.go
dnwick: Done

adapter/internal/adapter/adapter.go
sdkottegoda: The reason why we didn't want to have the go routine was to block the server startup if there is any issue with during connection initiation. But with the exsisting broker in a current deployment rollout, there'll be an almost definite round of reconnection attempts and AFAIR @Rajith90 mentioned this was changed to have a go routine because the server started hanging. 

Therefore I believe we can leave this as it it?

sdkottegoda (approved):

choreo-cicd:
[succeeded] dev-deployment : 20210911.3

PR #2375, opened 2021-11-03
Support env specific properties

AmaliMatharaarachchi:
### Issues

Fixes https://github.com/wso2/product-microgateway/issues/2389

### Automation tests
 - Unit tests added: Yes
 - Integration tests added: No

### Tested environments

Tested on catalina

---
- [x] Assigned 'Type' label
- [x] Assigned the project

### DIFF
--- a/adapter/internal/discovery/xds/server.go
+++ b/adapter/internal/discovery/xds/server.go
@@ -258,7 +258,7 @@ func UpdateAPI(vHost string, apiProject mgw.ProjectAPI, environments []string) (
 	}
 
 	if apiProject.APIType == mgw.HTTP || apiProject.APIType == mgw.WEBHOOK {
-		mgwSwagger, err = operator.GetMgwSwagger(apiProject.OpenAPIJsn)
+		mgwSwagger, err = operator.GetMgwSwagger(apiProject.OpenAPIJsn, apiProject.APIEnvProps)
 		if err != nil {
 			return nil, err
 		}
@@ -272,7 +272,7 @@ func UpdateAPI(vHost string, apiProject mgw.ProjectAPI, environments []string) (
 		mgwSwagger.SetXWso2AuthHeader(apiYaml.AuthorizationHeader)
 
 	} else if apiProject.APIType == mgw.WS {
-		mgwSwagger, err = operator.GetMgwSwaggerWebSocket(apiProject.APIYaml)
+		mgwSwagger, err = operator.GetMgwSwaggerWebSocket(apiProject.APIYaml, apiProject.APIEnvProps)
 		if err != nil {
 			return nil, err
 		}
--- a/adapter/internal/discovery/xds/server.go
+++ b/adapter/internal/discovery/xds/server.go
@@ -266,7 +266,7 @@ func UpdateAPI(vHost string, apiProject mgw.ProjectAPI, environments []string) (
 	}
 
 	if apiProject.APIType == mgw.HTTP || apiProject.APIType == mgw.WEBHOOK {
-		mgwSwagger, err = operator.GetMgwSwagger(apiProject.OpenAPIJsn, apiEnvProps)
+		mgwSwagger, err = operator.GetMgwSwagger(apiProject.OpenAPIJsn)
 		if err != nil {
 			return nil, err
 		}
@@ -280,7 +280,7 @@ func UpdateAPI(vHost string, apiProject mgw.ProjectAPI, environments []string) (
 		mgwSwagger.SetXWso2AuthHeader(apiYaml.AuthorizationHeader)
 
 	} else if apiProject.APIType == mgw.WS {
-		mgwSwagger, err = operator.GetMgwSwaggerWebSocket(apiProject.APIYaml, apiEnvProps)
+		mgwSwagger, err = operator.GetMgwSwaggerWebSocket(apiProject.APIYaml)
 		if err != nil {
 			return nil, err
 		}
@@ -288,6 +288,7 @@ func UpdateAPI(vHost string, apiProject mgw.ProjectAPI, environments []string) (
 		// Unreachable else condition. Added in case previous apiType check fails due to any modifications.
 		logger.LoggerXds.Error("API type not currently supported by Choreo Connect")
 	}
+	mgwSwagger.SetEnvProperties(apiEnvProps)
 	mgwSwagger.SetID(apiYaml.ID)
 	mgwSwagger.SetName(apiYaml.Name)
 	mgwSwagger.SetVersion(apiYaml.Version)
--- a/adapter/internal/discovery/xds/server.go
+++ b/adapter/internal/discovery/xds/server.go
@@ -48,6 +48,7 @@ import (
 	throttle "github.com/wso2/product-microgateway/adapter/pkg/discovery/api/wso2/discovery/throttle"
 	wso2_cache "github.com/wso2/product-microgateway/adapter/pkg/discovery/protocol/cache/v3"
 	eventhubTypes "github.com/wso2/product-microgateway/adapter/pkg/eventhub/types"
+	"github.com/wso2/product-microgateway/adapter/pkg/synchronizer"
 )
 
 var (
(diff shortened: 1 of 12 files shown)


adapter/internal/discovery/xds/server.go
```diff
@@ -272,7 +280,7 @@ func UpdateAPI(vHost string, apiProject mgw.ProjectAPI, environments []string) (
 		mgwSwagger.SetXWso2AuthHeader(apiYaml.AuthorizationHeader)
 
 	} else if apiProject.APIType == mgw.WS {
-		mgwSwagger, err = operator.GetMgwSwaggerWebSocket(apiProject.APIYaml)
+		mgwSwagger, err = operator.GetMgwSwaggerWebSocket(apiProject.APIYaml, apiEnvProps)
```
suksw, line 283: how about adding `mgwSwagger.SetEnvProperties(envProp)` after finishing both checks at 268 and 282? maybe after 293?

AmaliMatharaarachchi: yes. that's way better. thanks

adapter/internal/oasparser/envoyconf/routes_with_clusters_test.go
```diff
@@ -224,6 +228,50 @@ func testCreateRoutesWithClustersWebsocket(t *testing.T, apiYamlFilePath string)
...
+	envProps := synchronizer.APIEnvProps{
+		EnvID: "some id",
+		APIConfigs: synchronizer.APIConfigs{
+			ProductionEndpoint: "ws://env.websocket.org:443",
+			SandBoxEndpoint:    "ws://env.websocket.org:443",
+		},
+	}
+
+	apiYamlByteArr, err := ioutil.ReadFile(apiYamlFilePath)
+	assert.Nil(t, err, "Error while reading the api.yaml file : %v"+apiYamlFilePath)
```
renuka-fernando, line 242: Does "%v" in here is international?

adapter/pkg/synchronizer/apis_fetcher.go
```diff
@@ -223,20 +227,76 @@ func ReadDeployments(reader *zip.Reader) (*DeploymentDescriptor, error) {
...
+					var envProps APIEnvProps
+					for envLabel, envData := range api {
+						if err := parser.Decode(envData, &envProps); err != nil {
+							logger.LoggerSync.Debugf("Error parsing environment specific values: %v", err.Error())
+							return nil, err
+						}
+						apiProps[envLabel] = envProps
+					}
+				}
+				envProps[apiUUID] = apiProps
```
suksw, line 294: sorry for the delay after the first comment. Just thought of this. If this file sends data about many APIs, will there be a scenario we have to maintain them? As in, will there be an even where only the envs are sent without any APIs. Can that be the reason to send envProps that belongs to multiple APIs?

adapter/pkg/synchronizer/types.go
```diff
@@ -63,3 +63,15 @@ type GatewayLabel struct {
 	Name  string `json:"name"`
 	Vhost string `json:"vhost"`
 }
+
+// APIConfigs represents env properties belongs to the API
```
praminda, line 67: these structs are common to both synced and apict pushed APIs. so should be placed out of `synchronizer` pkg. `ProjectAPI` struct is in `oasparser` pkg. thats where this should be defined IMO

adapter/pkg/synchronizer/apis_fetcher.go
```diff
@@ -223,20 +227,76 @@ func ReadDeployments(reader *zip.Reader) (*DeploymentDescriptor, error) {
...
+				return deploymentDescriptor, nil, err
+			}
+			logger.LoggerSync.Debugf("Parsing content of environment specific properties, content: %s", string(fileContent))
+			apiEnvProps, err := parseEnvProps(fileContent)
+			if err != nil {
+				logger.LoggerSync.Errorf("Error occurred while parsing environment specific properties : %v : %v",
+					file.Name, err.Error())
+				return deploymentDescriptor, nil, err
+			} else if apiEnvProps != nil {
+				return deploymentDescriptor, apiEnvProps, nil
```
suksw, line 264: is it okay to return? we are still inside the loop right?

adapter/internal/oasparser/model/mgw_swagger.go
```diff
@@ -281,6 +282,42 @@ func (swagger *MgwSwagger) SetXWso2Extensions() error {
 	return nil
 }
 
+// SetEnvProperties sets environment specific values
+func (swagger *MgwSwagger) SetEnvProperties(envProps synchronizer.APIEnvProps) {
+	var productionUrls []Endpoint
+	var sandboxUrls []Endpoint
+
+	if envProps.APIConfigs.ProductionEndpoint != "" {
+		logger.LoggerOasparser.Infof("Production endpoints are found in env properties")
```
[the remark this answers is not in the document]
pubudu538: Should we log the API name as well?

adapter/pkg/synchronizer/apis_fetcher.go
pubudu538: This file is only coming from APIM internal rest API. This file contains values for all the APIs if they have some env values. That is the only requirement we have at the moment.

pubudu538 (commented):
Added some comments

adapter/pkg/synchronizer/types.go
```diff
@@ -63,3 +63,15 @@ type GatewayLabel struct {
...
+type APIConfigs struct {
+	ProductionEndpoint string `mapstructure:"production_endpoint,omitempty"`
```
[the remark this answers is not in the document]
AmaliMatharaarachchi: yes, it was recently updated to be only one url from APIM side. ack.

adapter/pkg/synchronizer/apis_fetcher.go
suksw: This would also mean that every time an API event occurs, we will be looping through all the APIs, for all its envs, and will be creating maps, unmarshelling objects, and decoding.

adapter/pkg/synchronizer/types.go
AmaliMatharaarachchi: This struct is used for the env_properties.json which is coming from APIM side. not supported for apictl at the moment. Hence keeping it in the synchronizer.

adapter/pkg/synchronizer/apis_fetcher.go
suksw: @AmaliMatharaarachchi how about using the method parseEnvProps after we get to know about the api uuid. then we can directly access the info we want and stop. the larger map can be discarded and only a smaller map will stay.

AmaliMatharaarachchi: @suksw The env_properties.json contains the environment label specific data only for the APIs corresponding to the API event. If I understand your concern correctly, so that won't be an issue since we only marshell env props of apis that we wants to update, am I right?

suksw: Got the idea after the discussion. So the env_properties.json is located outside the larger zip (which contains all the API zips). Therefore the file is meant to be used by multiple APIs. Thanks for the explanation. We can keep this as it is.

pubudu538 (approved):

adapter/internal/oasparser/envoyconf/routes_with_clusters_test.go
renuka-fernando: I think this is missed, isn't it?

renuka-fernando (approved):

choreo-cicd:
[failed] dev-deployment : 20211115.4

choreo-cicd:
[succeeded] dev-deployment : 20211115.12

choreo-cicd:
[failed] stage-deployment : 20211117.2

choreo-cicd:
[succeeded] stage-deployment : 20211117.3

choreo-cicd:
[succeeded] prod-deployment : 20211119.1

Issue #2363, opened 2021-11-01
Error un-deploying an API deployed to multiple labels form APICTL


renuka-fernando:
### Description:
Getting an error when un-deploying an API deployed to multiple labels through APICTL. The API is undeployed from one label. The error occurs when trying to un-deploy from the next label.

Adapter logs
```log
adapter_1     | 2021-11-01 10:08:59 INFO [auth.go:156] - [auth.ValidateToken] [-] Valid token recieved
adapter_1     | 2021-11-01 10:08:59 INFO [server.go:441] - [xds.DeleteAPIs] [-] Unable to delete API SwaggerPetstore:1.0.5 from Organization carbon.super. API does not exist.
```

APICTL
```sh
$ apictl mg undeploy api -n SwaggerPetstore -v 1.0.5 -e dev -k -g Default
apictl: Error undeploying API Reason: the API does not exist
Exit status 1
```

### Steps to reproduce:

Configure two gateway environments (Default and US).
- Default: one enforcer, one router
- US: one enforcer, one router
- adapter is common

Add the following config to the adapter
```toml
# Default virtual host mapping for standalone mode
[[adapter.vhostMapping]]
  # The Adapter environment
  environment = "Default"
  # Virtual host to map to the environment
  vhost = "localhost"
# Default virtual host mapping for standalone mode
[[adapter.vhostMapping]]
  # The Adapter environment
  environment = "US"
  # Virtual host to map to the environment
  vhost = "us.wso2.com"
```

Deploy the pet-store API to both labels. Used the following `deployment_environments.yaml` file.
```yaml
type: deployment_environments
version: v4.0.0
data:
 -
   displayOnDevportal: true
   deploymentEnvironment: US
 -
   displayOnDevportal: true
   deploymentEnvironment: Default
 -
   displayOnDevportal: true
   deploymentVhost: us.wso2.com
   deploymentEnvironment: Default
```

```
$ apictl mg deploy api -f petstore -e dev -k
Successfully deployed API to microgateway.
```

```
$ apictl mg get apis -e dev -k
Limit flag not set. Set to default: 25
APIs total: 2 received: 2
NAME                VERSION             TYPE                CONTEXT             GATEWAY_ENVS        VHOST
SwaggerPetstore     1.0.5               HTTP                /v2                 [US Default]        us.wso2.com
SwaggerPetstore     1.0.5               HTTP                /v2                 [Default]           localhost
```

```
[...]

PR #2489, opened 2021-11-26
[bug-fix] Fixes un-deploying an API deployed to multiple labels form APICTL

slahirucd7:
### Purpose
Fixes error when un-deploying an API deployed to multiple labels through APICTL.
image

### Issues

Fixes https://github.com/wso2/product-microgateway/issues/2363

### Automation tests
 - Unit tests added: No
 - Integration tests added: No

### Tested environments

---
- [x] Assigned 'Type' label
- [x] Assigned the project

### DIFF
--- a/adapter/internal/discovery/xds/server.go
+++ b/adapter/internal/discovery/xds/server.go
@@ -646,7 +646,7 @@ func deleteAPI(apiIdentifier string, environments []string, organizationID strin
     }
 
 	//clean maps of routes, clusters, endpoints, enforcerAPIs
-	if (len(environments) == 0) {
+	if len(environments) == 0 {
 		cleanMapResources(apiIdentifier, organizationID, toBeDelEnvs)
 	}
 	return nil
--- a/adapter/internal/discovery/xds/server.go
+++ b/adapter/internal/discovery/xds/server.go
@@ -502,6 +502,14 @@ func DeleteAPIs(vhost, apiName, version string, environments []string, organizat
 				return err
 			}
 			deletedVhosts[vh] = void
+			
+			for val := range deletedVhosts {
+				existingLabels := orgIDOpenAPIEnvoyMap[organizationID][apiIdentifier]
+				if val == vh && len(existingLabels) == 0 {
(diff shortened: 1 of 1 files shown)


pubudu538 (approved):

VirajSalaka:
Would be better if we can have an integration test to cover the scenario?
Something like this.
https://github.com/wso2/product-microgateway/commit/262838186350c855da83d5283189fe2883911d93

slahirucd7:
> Would be better if we can have an integration test to cover the scenario? Something like this. [2628381](https://github.com/wso2/product-microgateway/commit/262838186350c855da83d5283189fe2883911d93)

Ack. I'll add  a test with `Add integration tests for disable-security` PR.

adapter/internal/discovery/xds/server.go
```diff
@@ -502,8 +502,15 @@ func DeleteAPIs(vhost, apiName, version string, environments []string, organizat
...
 			}
 			deletedVhosts[vh] = void
+			
+			for val := range deletedVhosts {
+				existingLabels := orgIDOpenAPIEnvoyMap[organizationID][apiIdentifier]
+				if val == vh && len(existingLabels) == 0 {
+					logger.LoggerXds.Infof("Vhost : %v  deleted since there is no gateways assigned to it.", vh)
+					delete(apiToVhostsMap[apiNameVersionHashedID],val)
+				}
+			}
```
renuka-fernando, line 512: Can't we get this out from the parent for loop? And avoid checking `val == vh` in if condition?
Just a suggestion

slahirucd7: we need to check `val == vh` equality to determine deleting vhost properly. Also we need to do this inside the parent loop in order to delete the vhost at the time of gateway deletion (when there are no gateways allocated to the vhost vhost will also get deleted).

choreo-cicd:
[failed] dev-deployment : 20211215.32

choreo-cicd:
[succeeded] dev-deployment : 20211215.43

PR #2782, opened 2022-03-24
Masked the RevokedToken and log only with debug mode in revoked_token_listener.go

Amila-Rukshan:
### Purpose

Added a Maskstring util function in adapter and used it to mask (kept the last four characters unmasked) the RevokedToken prior to logging. RevokedToken logging statements changed to Debugf in revoked_token_listener.go.

In adapter, debug logged token will look like below
image

### Automation tests
 - Unit tests added: Yes
 - Integration tests added: No

### Tested environments

OS: Linux
Env: Docker

---
- [x] Assigned 'Type' label
- [x] Assigned the project


pubudu538 (approved):

adapter/internal/utils/string_utils.go
```diff
@@ -0,0 +1,40 @@
...
+	if 0 <= visibleLength && visibleLength < stringLength {
+		maskLength := stringLength - visibleLength
+		mask := strings.Repeat(maskCharacter, maskLength)
+		if maskRight {
+			return str[:visibleLength] + mask
+		}
+		return mask + str[maskLength:]
+	}
+
+	return str
```
VirajSalaka, line 39: Since this is a util function and method name is about masking, shall we not return string as it is? May be we can write it like, if visibleLength  > stringLength; print the mask character upto Visible Length. (This would be handy for masking shorter passwords event)

pubudu538: +1

VirajSalaka:
In the revokedTokenDiscoveryClient (within enforcer) also we log those information. Shall we stop printing that info at all?https://github.com/wso2/product-microgateway/blob/3eeca8a9d55efde8e0c4ace1721a7de85658d519/enforcer-parent/enforcer/src/main/java/org/wso2/choreo/connect/enforcer/discovery/RevokedTokenDiscoveryClient.java#L132

adapter/internal/utils/string_utils.go
```diff
@@ -0,0 +1,40 @@
+/*
+ *  Copyright (c) 2021, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
```
renuka-fernando, line 2:
```suggestion
 *  Copyright (c) 2022, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
```

adapter/internal/utils/string_utils.go
```diff
@@ -0,0 +1,40 @@
...
+)
+
+// MaskString takes string as argument and keep a visibleLength number of charators and mask rest of the string with maskCharacter
+// It can mask the charators from either left or right hand side of the string depending on maskRight boolean
+// It returns the masked string
+func MaskString(str string, visibleLength int, maskCharacter string, maskRight bool) string {
+	stringLength := len(str)
+
+	if 0 <= visibleLength && visibleLength < stringLength {
+		maskLength := stringLength - visibleLength
```
renuka-fernando, line 31: Do we want to have mask length the same as the actual length? @pubudu538
Just having three or ten of * is not enough? like `**********klm`

adapter/internal/utils/string_utils.go
```diff
@@ -0,0 +1,40 @@
...
+package stringutils
+
+import (
+	"strings"
+)
+
+// MaskString takes string as argument and keep a visibleLength number of charators and mask rest of the string with maskCharacter
+// It can mask the charators from either left or right hand side of the string depending on maskRight boolean
+// It returns the masked string
+func MaskString(str string, visibleLength int, maskCharacter string, maskRight bool) string {
```
renuka-fernando, line 27: Shall we create a new function `MaskToken` with some default values and it calls this function with that default values. i.e. MaskToken calls -> MaskString(token, 4, "*", false), So we can have kind of masking in other places.

VirajSalaka (approved):

renuka-fernando (approved):

pubudu538 (approved):

Issue #2797, opened 2022-03-26
[websocket] io.grpc.StatusRuntimeException and NPE thrown when closing a connection


suksw:
### Description:
The following gets thrown when a websocket client sends a close frame and closes its connection while the server also gracefully closes.

```
[docker-java-stream--1464580521] INFO Enforcer - STDOUT: [2022-03-25 09:57:36,460][][] ERROR - {org.wso2.choreo.connect.enforcer.websocket.WebSocketResponseObserver} - websocket frame service onError: io.grpc.StatusRuntimeException: CANCELLED: cancelled before receiving half close [severity:Default error_code:0]
[docker-java-stream--1464580521] INFO Enforcer - STDERR: Mar 25, 2022 9:57:36 AM io.grpc.internal.SerializingExecutor run
[docker-java-stream--1464580521] INFO Enforcer - STDERR: SEVERE: Exception while executing runnable io.grpc.internal.ServerImpl$JumpToApplicationThreadServerStreamListener$1Closed@15cc9a5f
[docker-java-stream--1464580521] INFO Enforcer - STDERR: java.lang.NullPointerException
[docker-java-stream--1464580521] INFO Enforcer - STDERR: 	at java.base/java.util.concurrent.ConcurrentHashMap.containsValue(Unknown Source)
[docker-java-stream--1464580521] INFO Enforcer - STDERR: 	at java.base/java.util.concurrent.ConcurrentHashMap.contains(Unknown Source)
[docker-java-stream--1464580521] INFO Enforcer - STDERR: 	at org.wso2.choreo.connect.enforcer.grpc.WebSocketFrameService.removeObserver(WebSocketFrameService.java:51)
[docker-java-stream--1464580521] INFO Enforcer - STDERR: 	at org.wso2.choreo.connect.enforcer.websocket.WebSocketResponseObserver.onError(WebSocketResponseObserver.java:131)
[docker-java-stream--1464580521] INFO Enforcer - STDERR: 	at io.grpc.stub.ServerCalls$StreamingServerCallHandler$StreamingServerCallListener.onCancel(ServerCalls.java:270)
[docker-java-stream--1464580521] INFO Enforcer - STDERR: 	at io.grpc.internal.ServerCallImpl$ServerStreamListenerImpl.closedInternal(ServerCallImpl.java:353)
[...]

PR #2798, opened 2022-03-26
Fix StatusRuntimeException cancelled before receiving half close

suksw:
### Purpose
$subject

### Issues
Fixes https://github.com/wso2/product-microgateway/issues/2797

### Automation tests
 - Unit tests added: Yes/No

### Tested environments

Not Tested

---
- [x] Assigned 'Type' label
- [x] Assigned the project

### DIFF
--- a/envoy-filters/mgw-source/filters/http/mgw-wasm-websocket/filter.cc
+++ b/envoy-filters/mgw-source/filters/http/mgw-wasm-websocket/filter.cc
@@ -318,6 +318,7 @@ void MgwWebSocketContext::onLog() { LOG_TRACE(std::string("onLog " + std::to_str
 void MgwWebSocketContext::onDelete() {
   LOG_TRACE(std::string("onDelete " + std::to_string(id())));
   this->stream_handler_->close();
+  this->stream_handler_->reset();
  }
 
 // Callback used by the handler to pass the throttle response received by the gRPC stream.
--- a/envoy-filters/mgw-source/filters/http/mgw-wasm-websocket/filter.cc
+++ b/envoy-filters/mgw-source/filters/http/mgw-wasm-websocket/filter.cc
@@ -317,7 +317,7 @@ void MgwWebSocketContext::onLog() { LOG_TRACE(std::string("onLog " + std::to_str
 
 void MgwWebSocketContext::onDelete() {
   LOG_TRACE(std::string("onDelete " + std::to_string(id())));
-  this->stream_handler_->reset();
+  this->stream_handler_->close();
  }
 
 // Callback used by the handler to pass the throttle response received by the gRPC stream.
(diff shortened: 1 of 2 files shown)


envoy-filters/mgw-source/filters/http/mgw-wasm-websocket/filter.cc
```diff
@@ -317,7 +317,7 @@ void MgwWebSocketContext::onLog() { LOG_TRACE(std::string("onLog " + std::to_str
 
 void MgwWebSocketContext::onDelete() {
   LOG_TRACE(std::string("onDelete " + std::to_string(id())));
-  this->stream_handler_->reset();
+  this->stream_handler_->close();
```
VirajSalaka, line 320: Isn't reset is the better option. Once the mgwContext is deleted we have no purpose in keeping the stream half closed and let enforcer send whatever it is left. When this triggers, the websocket connection is no longer available.

suksw: Without this change the error `io.grpc.StatusRuntimeException: CANCELLED: cancelled before receiving half close` keeps occurring for every successful websocket connections when they close. This is the case for all websocket tests we have in main branch.
- In [here](https://github.com/envoyproxy/envoy/blob/80d3c1750a18915f37fa65ca9d1bf3c26dd87d0b/test/extensions/filters/http/wasm/test_data/test_grpc_stream_cpp.cc#L51) they had only used `close()` which is why I thought it would be alright. 
- Also, according to the explanation [here](https://github.com/proxy-wasm/proxy-wasm-cpp-sdk/blob/master/docs/wasm_filter.md#reset) reset() I get the feeling that reset is trying to close something before it is ready. I am not 100% sure about this but this change seems to stop the error.

VirajSalaka: I have tested keeping close and enabling trace logs for wasm filter. It seems if the request is half close, the server is not closing the connection (it seems grpc library inside enforcer is also outdated a bit). Hence it causes the stream_handlers within wasm filter to accumulate which could lead to memory leak.

suksw: Can this accumilation be due to not calling `responseStreamObserver.onCompleted()` in https://github.com/wso2/product-microgateway/blob/2b8672d4d00d6fed01251579855d50d5ecec2095/enforcer-parent/enforcer/src/main/java/org/wso2/choreo/connect/enforcer/websocket/WebSocketResponseObserver.java#L135 ?

VirajSalaka: But we can't call it explicitly in onNext method right? Because termination is always initiated from the client. (wasm filter)

suksw: Tested with the directions on switching-on trace and looking for the log `handler destructed`. The grpc handler instance did not get destructed as you mentioned. After trying out the above and a few other ways, just found out that all we had to do was call both `close()` and `reset()`. Updated the PR accordingly.

VirajSalaka: great stuff. Reset after a half close does not violate the protocol it seems too.
+1

VirajSalaka (approved):

Issue #2831, opened 2022-03-31
Inconsistent behavior in API default version


VajiraPrabuddhaka:
### Description:
We have observed unexpected behavior with API default versions, which is when we have multiple versions of an API with one default version, some other versions' API calls seems going through the default version's endpoint. 

Sometimes, all versions' calls are going through the default version and sometimes only few versions.

Noted this is happening only when the API resource path is `/*`, i.e. not reproducing when we have a path.

### Steps to reproduce:

1. Create an API with few versions and set one version as default
2. Try invoking multiple versions and the default version
Note: Make sure API resource paths are /* (all resource paths) 

### Environment details (with versions):
- OS:
- Client:
- Env (Docker/K8s):

---

PR #2833, opened 2022-03-31
Bug fix: Default versioned APIs would be called when a versioned API is expected if it has the resource /*

VirajSalaka:
### Purpose
$subject

Solution:
To the routes array, all the default versioned apis should be in the tail. 
Hence always, 
/context/version pattern is matched prior to matching /context/*

### Issues

Fixes https://github.com/wso2/product-microgateway/issues/2831

### Automation tests
 - Unit tests added: No
 - Integration tests added: No

### Tested environments

Not Tested

---
- [x] Assigned 'Type' label
- [x] Assigned the project

### DIFF
--- a/adapter/internal/api/apis_impl.go
+++ b/adapter/internal/api/apis_impl.go
@@ -306,7 +306,6 @@ func ApplyAPIProjectFromAPIM(
 			// APIListMap is synchronously updated only for default version changes. In other API deployment
 			// events, this may not be updated. We can safely ignore this case since runtime artifact's
 			// `isDefaultVersion` prop is anyway updated for deployment events.
-			// TODO: (VirajSalaka) Is it better to set it to false?
 			loggers.LoggerAPI.Debugf("API %s is not found in API Metadata map.", apiYaml.ID)
 		}
 		// first update the API for vhost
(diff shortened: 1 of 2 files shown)


adapter/internal/api/apis_impl.go
```diff
@@ -306,6 +306,7 @@ func ApplyAPIProjectFromAPIM(
 			// APIListMap is synchronously updated only for default version changes. In other API deployment
 			// events, this may not be updated. We can safely ignore this case since runtime artifact's
 			// `isDefaultVersion` prop is anyway updated for deployment events.
+			// TODO: (VirajSalaka) Is it better to set it to false?
```
pubudu538, line 309:
```suggestion
			// TODO: (VirajSalaka) Is it better to set it to false?
```
```suggestion
			// TODO: (VirajSalaka) Is it better to set it to false?
```

pubudu538, line 309:
```suggestion
			// TODO: (VirajSalaka) Is it better to set it to false?
```
```suggestion
			// TODO: (VirajSalaka) Is it better to set it to false?
```

pubudu538, line 309:
```suggestion
			// TODO: (VirajSalaka) Is it better to set it to false?
```
```suggestion
```

pubudu538, line 309:
```suggestion
			// TODO: (VirajSalaka) Is it better to set it to false?
```
```suggestion
```

pubudu538 (approved):

Issue #2864, opened 2022-04-12
Removing "/" at the end of endpoint URL for API invocation cause 404 error.


IsuruBoyagane15:
### Description:
$subject

### Steps to reproduce:
- Try to invoke an API that is deployed to micro-gateway by removing the "/" at the end of the curl URL.

<img width="1488" alt="153913358-19c3e628-eaf6-4235-85aa-1438be1f5cca" src="https://user-images.githubusercontent.com/35162901/162887364-44c4444e-cbd9-4afa-8917-1b4172855f87.png">

### Environment details (with versions):
- OS:
- Client:
- Env (Docker/K8s):

---
#### Related Issues:

https://github.com/wso2-enterprise/choreo/issues/11341

PR #2837, opened 2022-04-01
Add trailing slash regex into path regex

IsuruBoyagane15:
### Purpose
To match both the paths with and without slash at the end of the URL when API is invoked.

### Issues

Fixes https://github.com/wso2-enterprise/choreo/issues/11341

### Automation tests
 - Unit tests added: Yes/No
 - Integration tests added: Yes/No

### Tested environments

 Ubuntu 20.04.4 , openjdk version "11.0.14.1" 

---
#### Maintainers: Check before merge
- [ ] Assigned 'Type' label
- [ ] Assigned the project
- [ ] Validated respective github issues
- [ ] Assigned milestone to the github issue(s)

PR #2878, opened 2022-04-20
[main] Update fix for trailing stash and add tests

suksw:
### Purpose
- A minor fix for a bug introduced when adding the change https://github.com/wso2/product-microgateway/pull/2837
- Add tests 
  - Invoking path with a trailing slash not given in the initial API definition
  - Invoke rewritten path (via API policy) with a trailing slash
  - Invoke default and non-default APIs with a trailing slash
  - Invoke `/context/*` for all methods when path is /*
  - Invoke `/context` for a few methods (get and post) when path is /*

### Issues
Related to https://github.com/wso2/product-microgateway/issues/2864

### Automation tests
 - Unit tests added: Updated

### Tested environments

Not Tested

---
- [x] Assigned 'Type' label
- [x] Assigned the project

### DIFF
--- a/adapter/internal/oasparser/envoyconf/routes_with_clusters.go
+++ b/adapter/internal/oasparser/envoyconf/routes_with_clusters.go
@@ -1390,6 +1390,8 @@ func getDefaultResourceMethods(apiType string) []string {
 }
 
 func getDefaultVersionBasepath(basePath string, version string) string {
-	context := strings.ReplaceAll(basePath, "/"+version, "")
+	// Following is used to replace only the version when basepath = /foo/v2 and version = v2 and context => /foo/v2/v2
+	indexOfVersionString := strings.LastIndex(basePath, "/"+version)
+	context := strings.Replace(basePath, "/"+version, "", indexOfVersionString)
 	return fmt.Sprintf("(?:%s|%s)", basePath, context)
 }
--- a/adapter/internal/oasparser/envoyconf/routes_with_clusters.go
+++ b/adapter/internal/oasparser/envoyconf/routes_with_clusters.go
@@ -1393,5 +1393,8 @@ func getDefaultVersionBasepath(basePath string, version string) string {
 	// Following is used to replace only the version when basepath = /foo/v2 and version = v2 and context => /foo/v2/v2
 	indexOfVersionString := strings.LastIndex(basePath, "/"+version)
 	context := strings.Replace(basePath, "/"+version, "", indexOfVersionString)
+
+	// Having ?: in the regex below, avoids this regex acting as a capturing group. Without this the basepath
+	// would again be added in the locations of path variables when sending the request to backend.
 	return fmt.Sprintf("(?:%s|%s)", basePath, context)
 }
--- a/adapter/internal/oasparser/envoyconf/routes_with_clusters.go
+++ b/adapter/internal/oasparser/envoyconf/routes_with_clusters.go
@@ -1391,5 +1391,5 @@ func getDefaultResourceMethods(apiType string) []string {
 
 func getDefaultVersionBasepath(basePath string, version string) string {
 	context := strings.ReplaceAll(basePath, "/"+version, "")
-	return fmt.Sprintf("(%s|%s)", basePath, context)
+	return fmt.Sprintf("(?:%s|%s)", basePath, context)
 }
--- a/integration/test-integration/src/test/java/org/wso2/choreo/connect/tests/testcases/standalone/apiDeploy/AllHttpMethodsForWildcardTestCase.java
+++ b/integration/test-integration/src/test/java/org/wso2/choreo/connect/tests/testcases/standalone/apiDeploy/AllHttpMethodsForWildcardTestCase.java
@@ -1,0 +1,142 @@
+/*
+ * Copyright (c) 2022, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
+ *
+ * WSO2 Inc. licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.wso2.choreo.connect.tests.testcases.standalone.apiDeploy;
+
+import com.github.dockerjava.zerodep.shaded.org.apache.hc.core5.http.HttpStatus;
+import io.netty.handler.codec.http.HttpHeaderNames;
+import org.testng.Assert;
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.Test;
+import org.wso2.choreo.connect.mockbackend.dto.EchoResponse;
+import org.wso2.choreo.connect.tests.util.HttpResponse;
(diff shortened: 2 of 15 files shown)


integration/test-integration/src/test/resources/openAPIs/trailing_slash_openAPI.yaml
```diff
@@ -0,0 +1,73 @@
+openapi: "3.0.0"
+info:
+  version: 1.0.0
+  title: Trailing Slash
+servers:
+  - url: http://mockBackend:2383/v2
+x-wso2-basePath: /trailing-slash
+paths:
+  /echo-full/findByStatus:
```
VirajSalaka, line 9: Shall we check with another resource with a trailing slash and see the behavior?
ex: `/foo/ `

adapter/internal/oasparser/envoyconf/routes_with_clusters.go
```diff
@@ -1391,5 +1391,5 @@ func getDefaultResourceMethods(apiType string) []string {
 
 func getDefaultVersionBasepath(basePath string, version string) string {
 	context := strings.ReplaceAll(basePath, "/"+version, "")
```
VirajSalaka, line 1393: Shall we fix this logic such that only the trailing part will be swapped?
I think we could use LastIndexOf and then pick a substring.

adapter/internal/oasparser/envoyconf/routes_with_clusters.go
```diff
@@ -1391,5 +1391,5 @@ func getDefaultResourceMethods(apiType string) []string {
...
-	return fmt.Sprintf("(%s|%s)", basePath, context)
+	return fmt.Sprintf("(?:%s|%s)", basePath, context)
```
VirajSalaka, line 1394: Can you please explain the purpose behind `?`

suksw: This is to make "(%s|%s)" behave as a non-capturing group in the regex. Behavior we saw where the basepath got placed again in the locations of the path parameters, gets resolved when we add this. I'll add a comment here.

adapter/internal/oasparser/envoyconf/routes_with_clusters.go
suksw: This is the `getDefaultVersionBasepath`. This is correct right?

integration/test-integration/src/test/resources/openAPIs/trailing_slash_openAPI.yaml
suksw: I updated policy and default API testcases as you suggested. They call other paths that are not `/echo-full`. Changes made to Default API testcase covers all three cases. https://github.com/wso2/product-microgateway/pull/2878/commits/fc9d1952fb8ad3d82c7fd091e0ff550193152fa6 Will these do?

integration/test-integration/src/test/java/org/wso2/choreo/connect/tests/testcases/standalone/apiDeploy/AllHttpMethodsForWildcardTestCase.java
```diff
@@ -0,0 +1,142 @@
...
+
+    @BeforeClass(description = "Get Prod token")
+    void start() throws Exception {
+        jwtTokenProd = TokenUtil.getJwtForPetstore(TestConstant.KEY_TYPE_PRODUCTION, null, false);
+        headers.put(HttpHeaderNames.AUTHORIZATION.toString(), "Bearer " + jwtTokenProd);
+    }
+
+    @Test(description = "test path with trailing slash for GET")
+    public void testTrailingSlashGET() throws Exception {
+        EchoResponse echoResponse = Utils.invokeEchoGet(API_CONTEXT, "/",
```
pubudu538, line 50: I guess this test is for invoking /api_context/. We need another test case for /api_context. There is no end slash.

pubudu538: In APIM if you create an API with /* resource then you can invoke this as /apicontext/ and /apicontext. Can't we achieve the same here?

VirajSalaka: we should fix it then

adapter/internal/oasparser/envoyconf/routes_with_clusters.go
VirajSalaka: Great. Thanks for the explaination.

adapter/internal/oasparser/envoyconf/routes_with_clusters.go
VirajSalaka: Let's assume the basePath is `/foo/v2` and version is `/v2`. So the complete path would be `/foo/v2/v2` in that case, v2 would be replaced twice right?
This is an edge case actually, but I have seen people doing that still.

integration/test-integration/src/test/java/org/wso2/choreo/connect/tests/testcases/standalone/apiDeploy/AllHttpMethodsForWildcardTestCase.java
suksw: This works. Let me update the tests. Didn't work while writing tests. Must have written the test wrong.

adapter/internal/oasparser/envoyconf/routes_with_clusters.go
suksw: +1

integration/test-integration/src/test/java/org/wso2/choreo/connect/tests/testcases/standalone/apiDeploy/AllHttpMethodsForWildcardTestCase.java
suksw: fixed with https://github.com/wso2/product-microgateway/pull/2878/commits/10bdb144f96c4a1ffcea84c43ab7803b61415a9a

adapter/internal/oasparser/envoyconf/routes_with_clusters.go
suksw: fixed with https://github.com/wso2/product-microgateway/pull/2878/commits/3fb46d23794f856dd8102d2b17b8172ab0f13a37

adapter/internal/oasparser/envoyconf/routes_with_clusters.go
suksw: Fixed with https://github.com/wso2/product-microgateway/pull/2878/commits/c9b1c53fe6698d6413e3057d3915dd31808bf1df

integration/test-integration/src/test/resources/openAPIs/trailing_slash_openAPI.yaml
VirajSalaka: The idea here is to like verify if the swagger resource has a trailing `/` and when we are forwarding it to the backend, whether we append that remaining slash.
As I saw, all the tests are like there is a resource with no slash and we are checking the both options. But nothing covers the pattern I tell right?

suksw: Thanks for the explanation. Got the idea. Will update.

suksw: fixed with https://github.com/wso2/product-microgateway/pull/2878/commits/dc1a2a713cc236684b6626c495fd64319604069a

VirajSalaka (approved):

PR #2934, opened 2022-05-20
Add the /apis endpoint call with expand false flag.

VirajSalaka:
### Purpose

Since the /apis endpoint call was a costly operation (due to API Manager DB operations), this was removed from the choreo branch. As a result of that, switching an API to Blocked Lifecycle state did not function. 
 With this PR, the /apis endpoint call will be added again. Hence the fore-mentioned limitation will be removed. 
 
 In this case, the /apis endpoint will be called with the query parameter `expand=false` so that it would return default APIs and blocked APIs only. This call is made during the startup. Then onwards, the blocked APIs and default APIs are identified using message events. 
 
In addition, the unnecessary properties within the protobuf files are removed too. 

### Issues

Fixes #

### Automation tests
 - Unit tests added: Yes/No

### Tested environments

Not Tested

---
- [ ] Assigned 'Type' label
- [ ] Assigned the project


adapter/internal/discovery/xds/marshaller.go
```diff
@@ -511,12 +510,13 @@ func MarshalAPIMetataAndReturnList(apiList *types.APIList, initialAPIUUIDListMap
 // DeleteAPIAndReturnList removes the API from internal maps and returns the marshalled API List.
 // If the apiUUID is not found in the internal map under the provided environment, then it would return a
 // nil value. Hence it is required to check if the return value is nil, prior to updating the XDS cache.
-func DeleteAPIAndReturnList(apiUUID, organizationUUID string, gatewayLabel string) *subscription.APIList {
+func DeleteAPIAndReturnList(apiUUID string, gatewayLabel string) *subscription.APIList {
 	if _, ok := APIListMap[gatewayLabel]; !ok {
 		logger.LoggerXds.Debugf("No API Metadata is available under gateway Environment : %s", gatewayLabel)
 		return nil
 	}
 	delete(APIListMap[gatewayLabel], apiUUID)
```
VirajSalaka, line 518: Check for apiUUID and return nil if it is not found.

adapter/internal/discovery/xds/marshaller.go
```diff
@@ -528,18 +528,40 @@ func MarshalAPIForLifeCycleChangeEventAndReturnList(apiUUID, status, gatewayLabe
...
+			LcState: status,
+		}
+	} else {
+		// If the update for existing API entry is not "BLOCKED" then the API can be removed from the list.
+		// when the API is unavailable in the api metadata list received in the enforcer, it would be treated as
+		// an unblocked API.
+		// But if the API is a default version, those APIs needs to be kept. But with the lifecycle state != BLOCKED.
+		if status != blockedStatus && !apiEntry.IsDefaultVersion {
+			return DeleteAPIAndReturnList(apiUUID, gatewayLabel)
+		}
```
VirajSalaka, line 551: Comment out the possibilities the code could reach this far

VirajSalaka:
closing the PR as https://github.com/wso2/product-microgateway/pull/2985 covers the same implementation but in more organized manner in terms of commits.

PR #2959, opened 2022-06-06
Edge proxy mode configuration added

### LINKED ISSUE #2957: x-forwarded-for header for router in edge proxy configuration
### Describe your problem(s)
Envoy by default passes the `x-forwarded-for` to the upstream, so the upstream services can process the header and get information about the originating ip address, network hops etc. But when the router is used as an edge proxy, it has to append the ip address of the immediate downstream which is typically a load balancer.  This is not added by the default configuration. 

### Describe your solution
Set `use_remote_address = true` when the router is used as an edge proxy. 
https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto#envoy-v3-api-field-extensions-filters-network-http-connection-manager-v3-httpconnectionmanager-use-remote-address

### How will you implement it
Add a new config field to the router config. 
---

NomadXD:
### Purpose
Envoy by default passes the x-forwarded-for to the upstream, so the upstream services can process the header and get information about the originating ip address, network hops etc. But when the router is used as an edge proxy, it has to append the ip address of the immediate downstream which is typically a load balancer. This is not added by the default configuration.

- `UseRemoteAddress = true` appends the ip address of the immediate downstream of the router to the `x-forward-for` header. Typically in edge proxy deployment, ip address of the load balancer will be added.
- `UseRemoteAddress = false` (default) passes the `x-forward-for` header to the upstream without appending the immediate downstream node ip address. 

### Issues

Fixes #2957 

### Automation tests
 - Unit tests added: No
 - Integration tests added: No

### Tested environments

Mac OS Monterey 12.3.1
Openjdk version 11.0.15
Docker

---
- [x] Assigned 'Type' label
- [x] Assigned the project

### DIFF
--- a/resources/conf/config-for-eventhub.toml
+++ b/resources/conf/config-for-eventhub.toml
@@ -10,6 +10,7 @@
   listenerPort = 9090
   securedListenerPort = 9095
   systemHost = "localhost"
+  edgeProxyMode = false
 
 [router.keystore]
   certPath = "/home/wso2/security/keystore/mg.pem"
--- a/resources/conf/config-for-eventhub.toml
+++ b/resources/conf/config-for-eventhub.toml
@@ -10,7 +10,7 @@
   listenerPort = 9090
   securedListenerPort = 9095
   systemHost = "localhost"
-  edgeProxyMode = false
+  useRemoteAddress = false
 
 [router.keystore]
   certPath = "/home/wso2/security/keystore/mg.pem"
--- a/resources/conf/config.toml
+++ b/resources/conf/config.toml
@@ -18,7 +18,7 @@
   listenerPort = 9090
   securedListenerPort = 9095
   systemHost = "localhost"
-  edgeProxyMode = false
+  useRemoteAddress = false
 
 [router.keystore]
   certPath = "/home/wso2/security/keystore/mg.pem"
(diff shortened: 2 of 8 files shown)


pubudu538 (approved):

resources/conf/config-for-eventhub.toml
```diff
@@ -10,6 +10,7 @@
   listenerPort = 9090
   securedListenerPort = 9095
   systemHost = "localhost"
+  useRemoteAddress = false
```
renuka-fernando, line 13: Shall we add this to the k8s artifact as well?

resources/conf/config.toml
```diff
@@ -18,6 +18,7 @@
   listenerPort = 9090
   securedListenerPort = 9095
   systemHost = "localhost"
+  useRemoteAddress = false
```
renuka-fernando, line 21: Shall we add this to the k8s artifact as well?

resources/conf/config-for-eventhub.toml
NomadXD: Config added to `config-toml-configmap-for-eventhub.yaml`

resources/conf/config.toml
NomadXD: Config added to `config-toml-configmap.yaml`

renuka-fernando (approved):

PR #2985, opened 2022-07-04
Add the /apis endpoint call with expand false flag

### LINKED ISSUE #2857: Avoid pulling all the API metadata from /internal/data/v1/apis endpoint
### Describe your problem(s)

Currently, the choreo connect calls `/internal/data/v1/apis` endpoint during startup and pull all the APIs. This is used to validate subscriptions, to identify default apis and to identify blocked APIs.

But for subscription validation, the choreo connect can retrieve the API UUID using the runtime artifact fetched from API manager. For default APIs and blocked APIs, a separate query parameter conditions are introduced and we can rely on that. Apart from that, it is not necessary to call the /apis endpoint for each API deploy event received.

### How will you implement it

---

VirajSalaka:
Purpose
Since the /apis endpoint call was a costly operation (due to API Manager DB operations), this was removed from the choreo branch. As a result of that, switching an API to Blocked Lifecycle state did not function.
With this PR, the /apis endpoint call will be added again. Hence the fore-mentioned limitation will be removed.

In this case, the /apis endpoint will be called with the query parameter expand=false so that it would return default APIs and blocked APIs only. This call is made during the startup. Then onwards, the blocked APIs and default APIs are identified using message events.

### Issues

Fixes #2857

### Automation tests
 - Unit tests added: No
 - Integration tests added: Yes (Covered by existing tests)

### Tested environments

Not Tested

---
- [ ] Assigned 'Type' label
- [ ] Assigned the project


adapter/internal/discovery/xds/server.go
```diff
@@ -1239,3 +1300,13 @@ func UpdateEnforcerThrottleData(throttleData *throttle.ThrottleData) {
...
 }
+
+func updateLCStateOfMgwSwagger(mgwSwagger *model.MgwSwagger) {
+	// If there are any metadata stored under the APIMetadataMap and Life Cycle state is blocked, update the mgwSwagger
+	apiEntry, apiFound := APIMetadataMap[mgwSwagger.GetID()]
+	if apiFound && apiEntry.LcState == blockedStatus {
+		mgwSwagger.LifeCycleState = blockedStatus
+		return
+	}
+	mgwSwagger.LifeCycleState = nonBlockedStatus
```
VirajSalaka, line 1311: Leave the lifecycleState as it is.

enforcer-parent/enforcer/src/main/java/org/wso2/choreo/connect/enforcer/admin/handlers/APIRequestHandler.java
```diff
@@ -82,10 +83,15 @@ private ResponsePayload getAPIs(String[] params) throws JsonProcessingException
...
             }
         }
-        apis = super.dataStore.getMatchingAPIs(name, context, version, uuid);
+        apis = APIFactory.getInstance().getMatchingAPIs(name, context, version, uuid);
         APIList apiList = new APIList();
         apiList.setCount(apis.size());
-        apiList.setList(apis);
+        List<BasicAPIInfo> modelAPIs = new ArrayList<>(apis.size());
+        for (API api : apis) {
+            // TODO: (VirajSalaka) fix
```
VirajSalaka, line 91: remove this todo

enforcer-parent/enforcer/src/main/java/org/wso2/choreo/connect/enforcer/models/API.java
```diff
@@ -20,117 +20,20 @@
 
 import org.wso2.choreo.connect.enforcer.common.CacheableEntity;
 
-import java.util.ArrayList;
```
VirajSalaka, line 23: Remove the API.java as it remains unused.

enforcer-parent/enforcer/src/main/java/org/wso2/choreo/connect/enforcer/models/admin/APIInfo.java
```diff
@@ -15,15 +15,14 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-package org.wso2.choreo.connect.enforcer.models;
```
VirajSalaka, line 18: Remove this class as it remains unused.

enforcer-parent/enforcer/src/main/java/org/wso2/choreo/connect/enforcer/subscription/SubscriptionDataStoreImpl.java
```diff
@@ -198,13 +196,6 @@ public void addApis(List<APIs> apisList) {
 
         for (APIs api : apisList) {
             API newApi = new API();
-            newApi.setApiId(Integer.parseInt(api.getApiId()));
```
VirajSalaka, line 201: We can remove this method as unused.

enforcer-parent/enforcer/src/main/java/org/wso2/choreo/connect/enforcer/subscription/SubscriptionDataStoreImpl.java
```diff
@@ -376,55 +367,16 @@ public void removeApiPolicy(ApiPolicy apiPolicy) {
...
-    @Override
-    public API getMatchingAPI(String context, String version) {
-        for (API api : apiMap.values()) {
-            if (StringUtils.isNotEmpty(context) && StringUtils.isNotEmpty(version)) {
-                if (api.getContext().equals(context) && api.getApiVersion().equals(version)) {
-                    return api;
-                }
-            }
-        }
+//        Set<String> set = apiMap.keySet()
```
VirajSalaka, line 370: Remove this as it remain unused.

enforcer-parent/enforcer/src/main/java/org/wso2/choreo/connect/enforcer/admin/AdminUtils.java
```diff
@@ -47,14 +48,28 @@ public class AdminUtils {
...
         apiInfo.setSubscriptions(subscriptionInfoList);
-        apiInfo.setApiId(api.getApiId());
-        apiInfo.setApiUUID(api.getApiUUID());
-        apiInfo.setContext(api.getContext());
-        apiInfo.setName(api.getApiName());
-        apiInfo.setLcState(api.getLcState());
-        apiInfo.setTier(api.getApiTier());
-        apiInfo.setVersion(api.getApiVersion());
-        apiInfo.setProvider(api.getApiProvider());
+        apiInfo.setApiUUID(api.getAPIConfig().getUuid());
```
VirajSalaka, line 51: Extends on BasicAPIInfo

adapter/internal/discovery/xds/server.go
VirajSalaka: Changed the lifecycle status to be published.

VirajSalaka: Reverted to previous state as we cannot decide whether the API is in created or published state when deployed for the first time. (Cannot depend on the API.yaml to populate this field)

enforcer-parent/enforcer/src/main/java/org/wso2/choreo/connect/enforcer/admin/AdminUtils.java
VirajSalaka: We had to remove tier as the BasicAPIInfo class already contains the policy. Since this is just used for troubleshooting I guess it is fine.

VirajSalaka:
There is no active progress being made here and /apis call is not used. And API Blocking feature is not working as a result of the initial fix and still the system remains same. If we need API blocking feature this need to be addressed in a separate way. 

Hence closing the PR.

PR #2992, opened 2022-07-07
Eliminate duplicate cluster creation when production and sandbox endpoints are the same

NomadXD:
### Purpose
Eliminate duplicate cluster creation when production and sandbox endpoints are the same

### Issues

Fixes #

### Automation tests
 - Unit tests added: Yes
 - Integration tests added: No

### Tested environments
OS: macOS Monterey 12.4
Env (Docker/K8s): Docker 20.10.16 on Rancher Desktop 1.4.1

---
- [ ] Assigned 'Type' label
- [ ] Assigned the project


pubudu538 (approved):

adapter/internal/oasparser/envoyconf/routes_with_clusters.go
```diff
@@ -182,37 +184,11 @@ func CreateRoutesWithClusters(mgwSwagger model.MgwSwagger, upstreamCerts map[str
...
-			logger.LoggerOasparser.Errorf("Error while adding api level response intercepter external cluster for %s. %v", apiTitle, err.Error())
-		} else {
-			clusters = append(clusters, cluster)
-			endpoints = append(endpoints, addresses...)
-		}
-	}
+	// Create API level interceptor clusters if required
+	clustersI, endpointsI, apiRequestInterceptor, apiResponseInterceptor := createInterceptorAPIClusters(mgwSwagger,
+		interceptorCerts, vHost, organizationID)
+	clusters = append(clusters, clustersI...)
```
AmaliMatharaarachchi, line 190: what happens when clustersI == nil i.e. when there's no interceptors?

AmaliMatharaarachchi: btw, can't we move this entire block out of the CreateRoutesWithClusters to a new method and refer that new method from GetRoutesClustersEndpoints method.

adapter/internal/oasparser/envoyconf/routes_with_clusters.go
```diff
@@ -1513,3 +1416,159 @@ func getDefaultVersionBasepath(basePath string, version string) string {
...
+		}
+		if productionEndpoint.Endpoints[0].Port != sandboxEndpoint.Endpoints[0].Port {
+			return true
+		}
+		// For Consul endpoints, check whether the service discovery strings are different.
+		if productionEndpoint.Endpoints[0].ServiceDiscoveryString != sandboxEndpoint.Endpoints[0].ServiceDiscoveryString {
+			return true
+		}
+	}
+	return false
```
AmaliMatharaarachchi, line 1437: can combine all conditions together

adapter/internal/oasparser/envoyconf/routes_with_clusters.go
```diff
@@ -141,21 +140,24 @@ func CreateRoutesWithClusters(mgwSwagger model.MgwSwagger, upstreamCerts map[str
...
+			apiLevelClusterNameSand = getClusterName(apiLevelEndpointSand.EndpointPrefix, organizationID, vHost,
+				apiTitle, apiVersion, "")
+			if !strings.Contains(apiLevelEndpointSand.EndpointPrefix, xWso2EPClustersConfigNamePrefix) {
+				cluster, address, err := processEndpoints(apiLevelClusterNameSand, apiLevelEndpointSand,
+					upstreamCerts, timeout, selectedBasePathSand)
+				if err != nil {
+					apiLevelClusterNameSand = ""
+					logger.LoggerOasparser.ErrorC(logging.ErrorDetails{
+						Message:   fmt.Sprintf("Error while adding api level sandbox endpoints for %s. %v", apiTitle, err.Error()),
+						Severity:  logging.CRITICAL,
```
VirajSalaka, line 154: IMO, this is not a critical error

VirajSalaka:
I guess it would be better to add an integration test too. 
An openapi with both production-endpoints and sandbox endpoints are the same. We do a call based on production token and sandbox token both. Covering the success path is enough.

adapter/internal/oasparser/envoyconf/routes_with_clusters.go
NomadXD: @AmaliMatharaarachchi 

1. `createInterceptorAPIClusters()` returns `[]*clusterv3.Cluster` and `[]*corev3.Address`. So if no intercepters that returned array of pointers will be empty and passed to the append method and nothing gets appended.

2. Yeah, I initially thought of moving the entire block out from `CreateRoutesWithClusters` but we pass request interceptors and response interceptors per resource to the `createRoute` method inside  `CreateRoutesWithClusters`. So if we are to completely take the interceptor creation code block out, we need to do following steps or something similar to that.

- Call `createInterceptorAPIClusters` from `GetRoutesClustersEndpoints` and save the interceptors in a map. We could probably have `resource:interceptor` mapping there.
- Pass the created interceptors to `CreateRoutesWithClusters` and in the for loop where we do `for _, resource := range mgwSwagger.GetResources()`, we can get the created interceptors using the resource key.

WDYT about that approach ?

AmaliMatharaarachchi: I suppose we are good with your above implementation then.

NomadXD: @AmaliMatharaarachchi Another approach is to separate cluster creation and route creation. We can split the logic inside `CreateRoutesWithClusters` into two.

1.   Create production, sandbox clusters, xwso2 clusters, API request and response interceptors first and store them in variables / maps.  (We are creating these per API not per resource)
2. In the next step we can iterate the resources and create resource and operational level interceptors if required and then finally call `createRoute` to create the route. We need to pass the clusters and API interceptors created in 1 and use them here. 

In this approach also we'll have to store the resources created in 1 and use them in 2.

I'll update this PR once the operational policies PR gets merged

pubudu538:
Shall we update this PR?

NomadXD:
@pubudu538 Sure, I'll update this with the proposed changes.

pubudu538 (approved):

VirajSalaka (approved):

Issue #3058, opened 2022-09-05
[int-tests] Add sub val failure test for validation via JWT claims only path for oauth and apikey


suksw:
### Description

$subject

Issue #2906, opened 2022-05-03
Subscription validation for API key is based on an irrelavant param


suksw:
### Description:
Currently, we enable subscription validation for **API key** based on the `validateSubscription` param in the config of the issuer "https://localhost:9443/publisher". During the time we added this change, the reason was to use a parameter that is separate from OAuth token configs. Yet, both API keys and OAuth tokens still has the same issuer (`https://localhost:9443/oauth2/token`) in its JWT payload.

In OAuth token validation we have the following scenarios. 
1. When sub validation for `https://localhost:9443/oauth2/token` is set to true,
  A. Validation of a token with API details (`subscribedApis` claim)
    - compares these details with the invoked API. Meant for running with keymanagers that are not aware about apps and for standalone mode.       
  
    B. Validation of a token that only has app details (created by resident key manager)
    - validates these **apps** together with the invoked API - > against the subscriptions details received from control plane.
  Note: When control plane is enabled, subscription validation for all issuers are always set to true. https://github.com/wso2/product-microgateway/blob/ea0651985b7dfbcbae9547304edfa72496ab4c2b/enforcer-parent/enforcer/src/main/java/org/wso2/choreo/connect/enforcer/keymgt/KeyManagerHolder.java#L108

2. When sub validation for `https://localhost:9443/oauth2/token` is set to `false`,
  - when control plane is enabled, updated back to `true` as given in the link above.
  - when control plane is disabled, any token, as long as signed by the correct private key, can invoke any API.

Currently, for API keys, 

3. when sub validation for "https://localhost:9443/publisher" is set true
- validation happens as 1. B.
- This is to make sure that subscription deletion and subscription blocking takes effect even if the API key doesn't expire. 
- Note: The content in the API key token is API-M resident key manager specific, unlike in aouth access tokens. Therefore, we enable subscription validation for **API-key** by default. 

4. when sub validation for "https://localhost:9443/publisher" is set false
- validate as 1. A.
- API-M creates api-keys that has a infinite validity period by default. To make such keys invalid, the cert the alias `gateway_certificate_alias` must be updated.
- therefore validating 1.A. way, atleast limits the api-key for one API.

### Suggested Fix
Instead of checking the issuer "https://localhost:9443/publisher", check if the control plane is enabled.
- if control plane is enabled -> validate subscriptions for API key like in 1.B. 
- else -> 1.A. 

### Environment details (with versions):
- OS:
- Client:
- Env (Docker/K8s):

---

[...]

PR #3077, opened 2022-09-16
Validate subscription in API key based on "APIM APIkey" issuer

suksw:
(this PR is based on changes suggested during discussions and review comments for 1.1.0.8)
### Purpose
- Fix certificate alias assignment and sub validation for API key and Internal key
  - Unlike oauth tokens, certs for API key and Internal key are not dynamic. They are configured during Choreo Connect startup. Hence, separating the dependency on the JWT ID in the header when picking the cert (alias). This would avoid having behaviors that are not visible in the config.
- Introduce a new issuer "APIM APIkey"
  ```
  [[enforcer.security.tokenService]]
    # Provide unique name for the JWT issuer
    name = "APIM APIkey"
    validateSubscription = true
    # Alias name given in Enforcer truststore for the public certificate of the JWT issuer
    certificateAlias = "apikey_certificate_alias"
    # Certificate Filepath within Enforcer
    certificateFilePath = "/home/wso2/security/truststore/wso2carbon.pem"
  ```
- For backward compatibility, we would still continue to depend on "APIM Publisher' issuer when "APIM APIkey" issuer is not present. "APIM Publisher" was originally used to refer to the Control Plane while maintaining a config for API key that was separate from the access token issuer (which was also Resident Key Manager).

  - "Resident Key manager" - Access tokens
  - "MGW"                 - Enforcer Test Key
  - "APIM APIkey"     - APIM API key
  - "APIM Publisher" - APIM Internal key (In case if "APIM APIkey" was not present will be used for API keys as well.)

- Access tokens, Internal Keys, API keys can have their own cert and cert alias for JWT signature validation.
- Disabling subscription validation for API key, will enable using same API key for any API as long as the signature is valid.

### Other Implementation details
- Add new param "ControlPlaneEnabled" to config proto. The reason is,
  - This location https://github.com/wso2/product-microgateway/blob/main/enforcer-parent/enforcer/src/main/java/org/wso2/choreo/connect/enforcer/keymgt/KeyManagerHolder.java#L108 set subValidation to true when control plane is enabled. 
  - Yet, there are two ways validate subscriptions.
    - when Control Plane enabled (withAPIM) -> check subscription store of enforcer.
    - when Control Plane disabled (standalone) -> check claims against invoked API 
- Update debug log to exclude query params.

Test changes - standalone
- Update existing apikey test to check sub validation failure for different API 
- Add apikey test with config including "APIM APIkey" issuer 
- Add test for oauth to check sub validation failure for different API  
- In CcWithMultipleEnv, remove the additional apictl env for better clarity of the gateway env
- Remove additional default configs from the test config multiple-env-config.toml

Test changes - withapim - none

### Issues
Fixes https://github.com/wso2/product-microgateway/issues/2906
Fixes https://github.com/wso2/product-microgateway/issues/3058

### Automation tests
 - Unit tests added: No

### Tested environments

Not Tested

---
- [x] Assigned 'Type' label
- [x] Assigned the project


VirajSalaka (approved):

enforcer-parent/enforcer/src/main/java/org/wso2/choreo/connect/enforcer/security/jwt/APIKeyAuthenticator.java
```diff
@@ -61,27 +62,43 @@
...
-            if (APIConstants.KeyManager.APIM_PUBLISHER_ISSUER.equals(tokenIssuer.getName())) {
+        for (ExtendedTokenIssuerDto tokenIssuer : enforcerConfig.getIssuersMap().values()) {
+            if (APIConstants.KeyManager.APIM_APIKEY_ISSUER.equals(tokenIssuer.getName())) {
+                certAlias = tokenIssuer.getCertificateAlias();
                 apiKeySubValidationEnabled = tokenIssuer.isValidateSubscriptions();
                 break;
             }
         }
+
+        // For backward compatibility
```
AmaliMatharaarachchi, line 88: do we require this for new cc releases?

adapter/config/default_config.go
```diff
@@ -180,7 +180,14 @@ var defaultConfig = &Config{
...
 					Issuer:               "https://localhost:9443/publisher",
 					ValidateSubscription: true,
-					CertificateAlias:     "",
+					CertificateAlias:     "publisher_certificate_alias",
+					CertificateFilePath:  "/home/wso2/security/truststore/wso2carbon.pem",
+				},
+				{
+					Name:                 "APIM APIkey",
+					Issuer:               "",
+					ValidateSubscription: true,
```
AmaliMatharaarachchi, line 189: The only use of this block is to enable or disable validation subscription for apim apikeys, right? I feel other fields in this config will not make much sense to a user
When cpenabled, are there any actual scenarios where users need to skip subscription validation? Does APIM too support turning off subscription validation for API keys?

suksw: The other scenarios would be to set a different certificate for API key. In that case, that cert would get stored under the alias defined in this block.

We enable subscription validation by default when CP is enabled. Did not come across a option to skip sub validation in APIM side. The only similar option was to disable security.

enforcer-parent/enforcer/src/main/java/org/wso2/choreo/connect/enforcer/security/jwt/APIKeyAuthenticator.java
pubudu538: Most of the case, we will be having the same use cases. So let's go with this for the moment. We can change these in the future if required.

adapter/config/default_config.go
AmaliMatharaarachchi: Does APIM support setting a separate cert for APIKeys? Isn't it issued by the same cert used for JWTs in APIM?

suksw: With the default configs, the Resident KM only sends a jwks endpoint. We retrive the cert when the initial cert arrives, and store the cert using the kid of the token as the alias. Unless we change the default configs, access tokens may have that can be updated time to time is what I got to know during a team discussion. And found this https://is.docs.wso2.com/en/latest/references/concepts/authentication/jwks/

AmaliMatharaarachchi (approved):

PR #3143, opened 2022-11-24
Attribute fix

SarithRavI:
### Purpose
Modification to retrieve the end user name and api context as a default analytics metric.

### Goals
By default, end user name and api context should be sent as a metric by the provider.

### Approach 
Fix the enforcer AnalyticsFilter
Fix the default analytics provider (ChoreoAnalyticsProvider)

### Automation tests
 - Unit tests added: No
 - Integration tests added: No

### Tested environments
JDK 11.0.16
Ubuntu 22.04.1 LTS

Not Tested


enforcer-parent/enforcer/pom.xml
```diff
@@ -634,4 +637,11 @@
         </dependency>
     </dependencies>
 
+    <properties>
+        <maven.compiler.source>11</maven.compiler.source>
+        <maven.compiler.target>11</maven.compiler.target>
+        <carbon.apimgt.common.analytics.version>9.27.17-SNAPSHOT</carbon.apimgt.common.analytics.version>
+        <org.wso2.am.analytics.publisher.version>1.1.1-SNAPSHOT</org.wso2.am.analytics.publisher.version>
```
tishan89, line 644: Please remove the versions from here as versioning will be fatched form the parent pom. Also always use released versions. Not SNAPSHOT version

enforcer-parent/enforcer/src/main/java/org/wso2/choreo/connect/enforcer/analytics/AnalyticsFilter.java
```diff
@@ -196,6 +196,13 @@ public void handleSuccessRequest(RequestContext requestContext) {
             requestContext.addMetadataToMap(MetadataConstants.CLIENT_IP_KEY, requestContext.getClientIp());
             requestContext.addMetadataToMap(MetadataConstants.USER_AGENT_KEY,
                     AnalyticsUtils.setDefaultIfNull(requestContext.getHeaders().get("user-agent")));
+
+            // Adding UserName and the APIContext
+            requestContext.addMetadataToMap(MetadataConstants.API_USER_NAME_KEY,
```
tishan89, line 201: What happens if username is null. Aren't we setting something like default user?

tishan89 (changes requested):

SarithRavI:
>> Please remove the versions from here as versioning will be fatched form
the parent pom. Also always use released versions. Not SNAPSHOT version
Ok Tishan..

>> What happens if username is null. Aren't we setting something like
default user?
Didn't occur that possibility for me. current implementation doesn't set a
default username, only set it to null if the username is unavailable.

enforcer-parent/enforcer/src/main/java/org/wso2/choreo/connect/enforcer/constants/MetadataConstants.java
```diff
@@ -29,6 +29,10 @@ public class MetadataConstants {
     public static final String API_CREATOR_KEY = WSO2_METADATA_PREFIX + "api-creator";
     public static final String API_NAME_KEY = WSO2_METADATA_PREFIX + "api-name";
     public static final String API_VERSION_KEY = WSO2_METADATA_PREFIX + "api-version";
+
+    public static final String API_USER_NAME_KEY = WSO2_METADATA_PREFIX + "user-name";
+
```
VirajSalaka, line 34: lets remove additional newlines

enforcer-parent/enforcer/src/main/java/org/wso2/choreo/connect/enforcer/analytics/AnalyticsFilter.java
VirajSalaka: This value could be null if the JWT token does not have a sub claim. In disabled security scenario we set it to "anonymous". may be we could use the same term here to handle username null scenario.

tishan89: shall we set as unknown

tishan89 (approved):

SarithRavI:
Closing this since necessary changes were made to the Choreo branch.

PR #3180, opened 2023-01-05
Adding a Proxy Server to Local Adapter

SarithRavI:
New proxy server has two types of downstreams.
 - Global Adapter
 - any external microservice(s)
 
 Currently only one microservice is added.
 One can add new microservice to the proxy by following similar steps to what discussed in wso2-enterprise/choreo-connect-global-adapter#243

### DIFF
--- a/adapter/config/types.go
+++ b/adapter/config/types.go
@@ -72,13 +72,15 @@ const (
 //	Don't use toml tag for configuration properties as it may affect environment variable based
 //	config resolution.
 type Config struct {
-	Adapter       adapter
-	Enforcer      enforcer
-	Envoy         envoy         `toml:"router"`
-	ControlPlane  controlPlane  `toml:"controlPlane"`
-	GlobalAdapter globalAdapter `toml:"globalAdapter"`
-	Analytics     analytics     `toml:"analytics"`
-	Tracing       tracing
+	Adapter            adapter
+	Enforcer           enforcer
+	Envoy              envoy         `toml:"router"`
+	ControlPlane       controlPlane  `toml:"controlPlane"`
+	GlobalAdapter      globalAdapter `toml:"globalAdapter"`
+	Analytics          analytics     `toml:"analytics"`
+	Tracing            tracing
+	MoesifMicroservice moesifMicroservice `toml:"moesifMicroservice"`
+	LAProxyServer      laProxyServer      `toml:"laProxyServer"`
 }
 
 // Adapter related Configurations
@@ -452,6 +454,19 @@ type tracing struct {
 	ConfigProperties map[string]string
 }
 
+type moesifMicroservice struct {
+	ServiceURL string `toml:"serviceUrl"`
+	Username   string `toml:"username"`
+	Password   string `toml:"password"`
+}
+
+type laProxyServer struct {
+	Host     string `toml:"host"`
+	Port     string `toml:"port"`
+	Username string `toml:"username"`
+	Password string `toml:"password"`
+}
+
 type metrics struct {
 	Enabled bool
 	Type    string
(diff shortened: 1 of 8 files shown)


adapter/config/types.go
```diff
@@ -452,6 +454,19 @@ type tracing struct {
 	ConfigProperties map[string]string
 }
 
+type moesifMicroservice struct {
+	ServiceURL string `toml:"serviceUrl"`
+	Username   string `toml:"username"`
+	Password   string `toml:"password"`
+}
```
VirajSalaka, line 461:
```suggestion
	ServiceURL string
	Username   string
	Password   string
}
```

adapter/config/types.go
```diff
@@ -452,6 +454,19 @@ type tracing struct {
...
+
+type laProxyServer struct {
+	Host     string `toml:"host"`
+	Port     string `toml:"port"`
+	Username string `toml:"username"`
+	Password string `toml:"password"`
```
VirajSalaka, line 467:
```suggestion
	Host     string
	Port     string
	Username string
	Password string
```

adapter/config/types.go
```diff
@@ -544,6 +559,8 @@ type globalAdapter struct {
 	ServiceURL string
 	// Deprecated: Use ServiceURL instead.
 	ServiceURLDeprecated string `toml:"serviceUrl"`
+	Username             string `toml:"username"`
+	Password             string `toml:"password"`
```
VirajSalaka, line 563: why we need to introduce these two properties?

adapter/internal/adapter/adapter.go
```diff
@@ -254,6 +254,12 @@ func Run(conf *config.Config) {
 		go restserver.StartRestServer(conf)
 	}
 
+	// Proxy server for the LA
+	if true {
```
VirajSalaka, line 258: This if condition has no meaning right?  Apart from that this is just a log only. I guess we need to start the server explicitly.

adapter/internal/proxyserver/handler.go
```diff
@@ -0,0 +1,217 @@
...
+	logger "github.com/wso2/product-microgateway/adapter/internal/loggers"
+)
+
+// HTTPGetToGAHandler function handles get requests
+func (s *Server) HTTPGetToGAHandler(w http.ResponseWriter, r *http.Request) {
+	buildRequestToGA(s.client, w, r, http.MethodGet, nil)
+}
+
+// HTTPPostToGAHandler function handles post requests
+func (s *Server) HTTPPostToGAHandler(w http.ResponseWriter, r *http.Request) {
```
VirajSalaka, line 41: We only need a single function right? let's remove everything else.

adapter/internal/proxyserver/handler.go
```diff
@@ -0,0 +1,217 @@
...
+	if resp.StatusCode != 200 {
+		bodyBytes, _ := ioutil.ReadAll(resp.Body)
+		bodyString := string(bodyBytes)
+		w.WriteHeader(resp.StatusCode)
+		fmt.Fprintf(w, "%s", bodyString)
+	}
+
+	bodyBytes, _ := ioutil.ReadAll(resp.Body)
+	bodyString := string(bodyBytes)
+	fmt.Fprintf(w, "%s", bodyString)
```
VirajSalaka, line 138: This is not relevant to the scope of this PR, I think we should need some sort of a retry mechanism and probably it would be better to introduce that at the enforcer level. Just for your information.

adapter/internal/adapter/adapter.go
SarithRavI: Yes Viraj..
This is a mistake.

adapter/internal/proxyserver/handler.go
SarithRavI: Ack.
I will keep only the GET resource.

adapter/internal/proxyserver/handler.go
SarithRavI: Ack.

SarithRavI:
Adding proxy server to local adapter is decided to be out of project's scope.

PR #3217, opened 2023-01-26
Attribute fix choreo

SarithRavI:
Please note that modifications in this PR are same as the **approved** PR (in main branch) #3143. 
### Purpose

Modification to retrieve the end user name and api context as a default analytics metric. 

### Goals
By default, end user name and api context should be sent as a metric by the provider.

### Approach
Fix the enforcer AnalyticsFilter
Fix the default analytics provider (ChoreoAnalyticsProvider)

### Automation tests
Unit tests added: No
 Integration tests added: No

### Tested environments

JDK 11.0.16
Ubuntu 22.04.1 LTS


SarithRavI:
Faulty PR.

PR #3363, opened 2023-05-28
[choreo] Include Environment name in GA API event

VajiraPrabuddhaka:
### Purpose

- This will introduce a new attribute Environment name into the GA API event. This attribute will be important when deploying/ undeploying APIs via Global Adapter with dynamic environment support feature is enabled.

- There seems to have an issue in the current flow when the GA is enabled. We need to wait until the initial GA API list is proceed, before deploying APIs with events. This PR will also address that issue.

### Issues

Fixes #

### Automation tests
 - Unit tests added: Yes/No

### Tested environments

Tested locally using a GA running locally by pointing to the dev environment.

---
- [x] Assigned 'Type' label
- [x] Assigned the project


api/proto/wso2/discovery/ga/api.proto
```diff
@@ -14,4 +14,5 @@ message Api {
   string apiUUID = 1;
   string revisionUUID = 2;
   string organizationUUID = 3;
+  string deployedEnv = 4;
```
VirajSalaka, line 17: should deployedEnv be optional? (If it is not breaking the current implementation, it is okay to keep this field as it is IMO)

adapter/internal/ga/client.go
```diff
@@ -155,7 +156,11 @@ func watchAPIs() {
 		} else {
 			lastReceivedResponse = discoveryResponse
 			logger.LoggerGA.Debugf("Discovery response is received : %s", discoveryResponse.VersionInfo)
-			addAPIToChannel(discoveryResponse)
+			if conf.GlobalAdapter.Enabled && conf.ControlPlane.DynamicEnvironments.Enabled {
```
VirajSalaka, line 159: `conf.GlobalAdapter.Enabled` remains always true as GA client never starts otherwise.

adapter/internal/ga/client.go
```diff
@@ -155,7 +156,11 @@ func watchAPIs() {
...
+				addAPIWithEnvToChannel(discoveryResponse)
+			} else {
+				addAPIToChannel(discoveryResponse)
```
VirajSalaka, line 162: Shall we keep a TODO statement here and then delete these unnecessary changes once the fix is rolled out.

adapter/internal/ga/client.go
```diff
@@ -46,7 +44,8 @@ import (
 var (
 	// apiRevision Map keeps apiUUID -> revisionUUID. This is used only for the communication between global adapter and adapter
 	// The purpose here is to identify if the certain API's revision is already added to the XDS cache.
-	apiRevisionMap map[string]*ga_model.Api
+	apiRevisionMap    map[string]*ga_model.Api
+	apiEnvRevisionMap = make(map[string]map[string]*ga_model.Api)
```
VirajSalaka, line 48: Let's add a comment just like in line 45 to indicate what is first key, and second key and finally the value.

adapter/internal/ga/client.go
```diff
@@ -288,13 +290,98 @@ func addAPIToChannel(resp *discovery.DiscoveryResponse) {
...
 	}
 }
 
+func addAPIWithEnvToChannel(resp *discovery.DiscoveryResponse) {
+
+	// removedAPIEnvMap is used To keep track of the APIs needs to be deleted.
+	var removedAPIEnvMap = make(map[string]map[string]*ga_model.Api)
+
+	if !isFirstResponse {
+		for k, v := range apiEnvRevisionMap {
```
VirajSalaka, line 307: shall we have meaningful terms for single character variable names.

api/proto/wso2/discovery/ga/api.proto
VajiraPrabuddhaka: I have locally tested this and seems it is optional by default. I tested this on a GA running with the older version (without having this field added) and a LA with the new version (with having this field added) and it was working without any issue. An empty value was received for the newly added field.

Also from the discussions [1] & [2] I think this behavior is correct for proto3.

[1] : https://stackoverflow.com/questions/31801257/why-required-and-optional-is-removed-in-protocol-buffers-3
[2] : https://github.com/protocolbuffers/protobuf/issues/2497#issuecomment-267422550

adapter/internal/ga/client.go
```diff
@@ -288,13 +290,98 @@ func addAPIToChannel(resp *discovery.DiscoveryResponse) {
...
+	for _, res := range resp.Resources {
+		api := &ga_model.Api{}
+		err := ptypes.UnmarshalAny(res, api)
+
+		if err != nil {
+			logger.LoggerGA.Errorf("Error while unmarshalling: %s\n", err.Error())
+			continue
+		}
+
+		currentGAAPI, apiFound := apiEnvRevisionMap[api.ApiUUID][api.DeployedEnv]
```
VirajSalaka, line 328: Let's assume that it is a newly created API, then it would end up with NPE

VirajSalaka: where apiEnvRevisionMap[api.ApiUUID] == nil

adapter/internal/ga/client.go
```diff
@@ -288,13 +290,98 @@ func addAPIToChannel(resp *discovery.DiscoveryResponse) {
...
+		// it appends to an array and let the apis_fetcher collect those data.
+		if isFirstResponse {
+			startupAPIEventArray = append(startupAPIEventArray, &event)
+		} else {
+			GAAPIChannel <- event
+		}
+		if _, found := apiEnvRevisionMap[api.ApiUUID]; !found {
+			apiEnvRevisionMap[api.ApiUUID] = make(map[string]*ga_model.Api)
+		}
+		apiEnvRevisionMap[api.ApiUUID][api.DeployedEnv] = api
```
VirajSalaka, line 355: NPE possibility is here.

adapter/internal/ga/client.go
```diff
@@ -77,6 +76,7 @@ type APIEvent struct {
 	RevisionUUID     string
 	IsDeployEvent    bool
 	OrganizationUUID string
+	DeployedEnv      string
```
VirajSalaka, line 79: what is deployedEnv ?  is it a UUID or a name? If it is a name, can we guarantee that it is unique for a given org. (Since each project can have its own env, there is a possibility two projects can have the same env name)

adapter/internal/ga/client.go
```diff
@@ -288,13 +290,98 @@ func addAPIToChannel(resp *discovery.DiscoveryResponse) {
...
+			logger.LoggerGA.Errorf("Error while unmarshalling: %s\n", err.Error())
+			continue
+		}
+
+		currentGAAPI, apiFound := apiEnvRevisionMap[api.ApiUUID][api.DeployedEnv]
+		if apiFound {
+			if currentGAAPI.RevisionUUID == api.RevisionUUID {
+				logger.LoggerGA.Debugf("Current GA API revision ID and API event revision ID: %s in "+
+					"environment: %s are equal", currentGAAPI.RevisionUUID, currentGAAPI.DeployedEnv)
+				delete(removedAPIEnvMap[api.ApiUUID], api.DeployedEnv)
```
VirajSalaka, line 333: Possible NPE if `removedAPIEnvMap` does not contain api.APIUUID.

VirajSalaka: In a perfect scenario, this won't be null but due to bug let's say per API, per Env there would be two deployments and isFirstResponse is true this could occur. 
Let's make sure code does not break avoiding the NPE.

adapter/internal/ga/client.go
```diff
@@ -288,13 +290,98 @@ func addAPIToChannel(resp *discovery.DiscoveryResponse) {
...
+		for envName, gaAPI := range envAPI {
+			event := APIEvent{
+				APIUUID:          apiID,
+				IsDeployEvent:    false,
+				OrganizationUUID: gaAPI.OrganizationUUID,
+				RevisionUUID:     gaAPI.RevisionUUID,
+				DeployedEnv:      envName,
+			}
+			GAAPIChannel <- event
+			delete(apiEnvRevisionMap[apiID], envName)
```
VirajSalaka, line 377: Let's avoid possible NPE

adapter/internal/ga/client.go
VajiraPrabuddhaka: No, there is no NPE here. What happens is `apiFound` becomes `false` and `currentGAAPI` becomes nil. I have verified this scenario in locally debugging.

adapter/internal/ga/client.go
VajiraPrabuddhaka: Here we are using the env name along with APIUUID and RevisionUUID. So I think APIUUD, RevisionUUID, DeployedEnv all together brings that uniqueness here.

adapter/internal/ga/client.go
VajiraPrabuddhaka: fixed via https://github.com/wso2/product-microgateway/commit/ae2c06882353bda7a59101381ef3e82f88667a21

adapter/internal/ga/client.go
VajiraPrabuddhaka: addressed via https://github.com/wso2/product-microgateway/commit/ae2c06882353bda7a59101381ef3e82f88667a21

adapter/internal/ga/client.go
VajiraPrabuddhaka: addressed via https://github.com/wso2/product-microgateway/commit/ae2c06882353bda7a59101381ef3e82f88667a21

adapter/internal/ga/client.go
VajiraPrabuddhaka: fixed via https://github.com/wso2/product-microgateway/commit/ae2c06882353bda7a59101381ef3e82f88667a21

adapter/internal/ga/client.go
VajiraPrabuddhaka: add a check prior to deleting https://github.com/wso2/product-microgateway/commit/ae2c06882353bda7a59101381ef3e82f88667a21

adapter/internal/ga/client.go
VajiraPrabuddhaka: added a check prior to deletion https://github.com/wso2/product-microgateway/commit/ae2c06882353bda7a59101381ef3e82f88667a21

adapter/internal/ga/client.go
VirajSalaka: APIUUID is a unique one, and since the API belongs to exactly one project I think your argument stands.

adapter/internal/ga/client.go
VirajSalaka: Oh it seems golang handles it own its own.

VirajSalaka (approved):

VirajSalaka (approved):

choreo-cicd:
[succeeded] Dataplane(EastUS) cluster : dev-deployment-v2 : 20230529.16

choreo-cicd:
[succeeded] Dataplane(NorthEU) cluster : dev-deployment-v2 : 20230529.16

choreo-cicd:
[succeeded] Controlplane cluster : dev-deployment-v2 : 20230529.16

choreo-cicd:
[succeeded] Dataplane(EastUS) cluster : stage-deployment-v2 : 20230530.3

choreo-cicd:
[succeeded] Dataplane(NorthEU) cluster : stage-deployment-v2 : 20230530.3

choreo-cicd:
[succeeded] Controlplane cluster : stage-deployment-v2 : 20230530.3

choreo-cicd:
[succeeded] Dataplane(NorthEU) cluster : prod-deployment-v2 : 20230530.3

PR #3376, opened 2023-06-12
Implement semantic version based intelligent routing for APIs

chathuranga95:
### Purpose
This PR adds intelligent routing implementations for APIs using semantic versioning.

**Added routing patterns**
Consider an API with multiple versions. Eg: 1.0.0, 1.0.1, 1.0.2, 1.2.0, 2.0.0
1. `/foo/1.0.0` ---> invokes exactly the 1.0.0 version. applies for all APIs. This pattern ensures all existing patterns are matched
2. `/foo/v1` ---> Invokes the latest semantic version of the major version: 1 range. Eg: 1.2.0
3. `/foo/v1.0` ---> Invokes the latest semantic version of the minor version: 1.0 range. Eg: 1.0.2

`/foo/v1.0.0` is not added as a supported pattern in order to maintain consistency and differentiate intelligent routing and exact version based invocations.

> Note: Don't merge until the other related tasks from the original epic are finalized and completed.

### Automation tests
 - Unit tests added: TBA
 - Integration tests added: TBD

### Tested environments

OS: ubuntu 20.04
Env: docker

---
- [x] Assigned 'Type' label
- [x] Assigned the project


adapter/internal/discovery/xds/server.go
```diff
@@ -251,6 +256,55 @@ func DeployReadinessAPI(envs []string) {
 	}
 }
 
+// ValidateAndGetVersionComponents validates version string and extracts version components
+func ValidateAndGetVersionComponents(version string) (string, string, string, error) {
+	versionComponents := strings.Split(version, ".")
+
+	// If the versionComponents length is less than 3, return error
```
renuka-fernando, line 263:
```suggestion
	// If the versionComponents length is less than 2, return error
```

adapter/internal/discovery/xds/server.go
```diff
@@ -466,6 +520,109 @@ func UpdateAPI(vHost string, apiProject mgw.ProjectAPI, deployedEnvironments []*
 		orgIDOpenAPIRoutesMap[organizationID] = routesMap
 	}
 
+	apiName := mgwSwagger.GetTitle()
+	apiVersion := mgwSwagger.GetVersion()
+	apiRangeIdentifier := GenerateIdentifierForAPIWithoutVersion(vHost, apiName)
+
+	_, _, _, err = ValidateAndGetVersionComponents(apiVersion)
```
renuka-fernando, line 527: Shall we do this inside the if block `if isIntelligentRoutingEnabled` and reuse the returned values instead of calling it multiple times?

renuka-fernando: Maybe we can return a struct pointer from this function. For example. And we can reuse this value in other functions.

```go
type struct SemVersion {
	Major int
	Minor int
	Patch int
}

func ValidateAndGetVersionComponents(version string) (*SemVersion, error) {
}
```

adapter/internal/discovery/xds/server.go
```diff
@@ -466,6 +520,109 @@ func UpdateAPI(vHost string, apiProject mgw.ProjectAPI, deployedEnvironments []*
...
+	isIntelligentRoutingEnabled := conf.Adapter.IsIntelligentRoutingEnabled
+	if isIntelligentRoutingEnabled && err == nil {
+		// Check the major and minor version ranges of the current API
+		existingMajorRangeLatestVersion, isMajorRangeRegexAvailable :=
+			orgIDLatestAPIVersionMap[organizationID][apiRangeIdentifier][GetMajorVersionRange(apiVersion)]
+		existingMinorRangeLatestVersion, isMinorRangeRegexAvailable :=
+			orgIDLatestAPIVersionMap[organizationID][apiRangeIdentifier][GetMinorVersionRange(apiVersion)]
+
+		// Check whether the current API is the latest version in the major and minor version ranges
+		isLatestMajorVersion := !isMajorRangeRegexAvailable || existingMajorRangeLatestVersion <= apiVersion
```
renuka-fernando, line 538: `existingMajorRangeLatestVersion <= apiVersion` Is this a string comparison?
I mean what if existingMajorRangeLatestVersion is 12 and apiVersion is 2, 12 <= 2

adapter/internal/discovery/xds/server.go
```diff
@@ -466,6 +520,109 @@ func UpdateAPI(vHost string, apiProject mgw.ProjectAPI, deployedEnvironments []*
...
+
+		if isLatestMajorVersion || isLatestMinorVersion {
+			// Update local memory map with the latest version ranges
+			if _, ok := orgIDLatestAPIVersionMap[organizationID]; !ok {
+				latestVersions := make(map[string]string)
+				latestVersions[GetMajorVersionRange(apiVersion)] = apiVersion
+				latestVersions[GetMinorVersionRange(apiVersion)] = apiVersion
+				apiVersionMap := make(map[string]map[string]string)
+				apiVersionMap[apiRangeIdentifier] = latestVersions
+				orgIDLatestAPIVersionMap[organizationID] = apiVersionMap
```
renuka-fernando, line 586: When APIs are getting undeployed, doesn't the entries in this map is removed?

adapter/internal/discovery/xds/server.go
chathuranga95: > Shall we do this inside the if block `if isIntelligentRoutingEnabled` and reuse the returned values instead of calling it multiple times?

+1 to move this inside `if isIntelligentRoutingEnabled` block and have another if block to check whether the version is compatible with semantic versioning.
Note that the `ValidateAndGetVersionComponents` function is used on different versions such as current api version, current latest major range version and current latest minor range version. Therefore, we can't reuse the returned values.

chathuranga95: > Maybe we can return a struct pointer from this function. For example. And we can reuse this value in other functions.
> 
> ```go
> type struct SemVersion {
> 	Major int
> 	Minor int
> 	Patch int
> }
> 
> func ValidateAndGetVersionComponents(version string) (*SemVersion, error) {
> }
> ```

+1, with Patch is being an optional field

adapter/internal/discovery/xds/server.go
chathuranga95: Will fix by using integer components as suggested above

adapter/internal/discovery/xds/server.go
renuka-fernando: Got your point. What if we change the type of the internal memory map to the following?
```go
orgIDLatestAPIVersionMap map[string]map[string]map[string]SemVersion
```
So we do not need to recompute this value again.

adapter/internal/discovery/xds/server.go
renuka-fernando: Eg:

When `1.2.0` is undeployed.
`/foo/v1` ---> 1.2.0 should be changed to `/foo/v1` ---> 1.0.2

adapter/internal/discovery/xds/server.go
chathuranga95: This will introduce another overhead of concatenating and building the version from the version components again.

We may be able to overcome these computing overheads by updating the `SemVersion` to hold the exact version and the version components as well.

adapter/internal/discovery/xds/server.go
chathuranga95: Fixed with: https://github.com/wso2/product-microgateway/pull/3376/commits/8a1b6d61e5c8147d8bc7ed747b353ae474f90d05

adapter/internal/discovery/xds/server.go
chathuranga95: Fixed with https://github.com/wso2/product-microgateway/pull/3376/commits/bc8484ac9a3f5629075cd867af5e0659ecf3b80f, https://github.com/wso2/product-microgateway/pull/3376/commits/f1b404e5a34f6b004ac91f88dd4af63ff7bf9644 and https://github.com/wso2/product-microgateway/pull/3376/commits/10575c0498ec6037c245cc1473891adaad8ee27a

adapter/internal/discovery/xds/server.go
chathuranga95: Fixed with: https://github.com/wso2/product-microgateway/commit/10575c0498ec6037c245cc1473891adaad8ee27a

adapter/internal/discovery/xds/server.go
chathuranga95: Added routing updating on API undeployment with: https://github.com/wso2/product-microgateway/pull/3376/commits/a7a452c60f9eddc276ecb9713e1707cb9d9415cb

adapter/internal/discovery/xds/server.go
```diff
@@ -466,6 +478,110 @@ func UpdateAPI(vHost string, apiProject mgw.ProjectAPI, deployedEnvironments []*
 		orgIDOpenAPIRoutesMap[organizationID] = routesMap
 	}
 
+	apiName := mgwSwagger.GetTitle()
+	apiVersion := mgwSwagger.GetVersion()
+	apiRangeIdentifier := GenerateIdentifierForAPIWithoutVersion(vHost, apiName)
+
+	isIntelligentRoutingEnabled := conf.Adapter.IsIntelligentRoutingEnabled
```
slahirucd7, line 485: Can change as below ?
if `conf.Adapter.IsIntelligentRoutingEnabled`

adapter/internal/discovery/xds/server_utils.go
```diff
@@ -81,3 +85,110 @@ func updateVhostInternalMaps(uuid, name, version, vHost string, gwEnvs []string)
...
+	} else if baseVersion.Major > version.Major {
+		return false
+	} else {
+		if baseVersion.Minor < version.Minor {
+			return true
+		} else if baseVersion.Minor > version.Minor {
+			return false
+		} else {
+			if baseVersion.Patch != nil && version.Patch != nil {
+				if *baseVersion.Patch < *version.Patch {
```
slahirucd7, line 181: Can we use `baseVersion.Patch` in here as well ?

adapter/internal/discovery/xds/server.go
```diff
@@ -466,6 +478,110 @@ func UpdateAPI(vHost string, apiProject mgw.ProjectAPI, deployedEnvironments []*
...
+			isLatestMajorVersion := !isMajorRangeRegexAvailable || CompareSemanticVersions(existingMajorRangeLatestSemVersion, *apiSemVersion)
+			isLatestMinorVersion := !isMinorRangeRegexAvailable || CompareSemanticVersions(existingMinorRangeLatestSemVersion, *apiSemVersion)
+
+			// Remove the existing regexes from the path specifier when latest major and/or minor version is available
+			if (isMajorRangeRegexAvailable || isMinorRangeRegexAvailable) && (isLatestMajorVersion || isLatestMinorVersion) {
+				// Organization's all apis
+				for apiUUID, swagger := range orgIDAPIMgwSwaggerMap[organizationID] {
+					// API's all versions in the same vHost
+					if swagger.GetTitle() == apiName && swagger.GetVHost() == vHost {
+						if isMajorRangeRegexAvailable && swagger.GetVersion() == existingMajorRangeLatestSemVersion.Version ||
```
slahirucd7, line 505: Can we add the 505 `if` conditions to the to the line 504 if conditions ?
or else what if we move the condition checking to a separate method ?

adapter/internal/discovery/xds/server.go
```diff
@@ -666,6 +782,137 @@ func deleteAPI(apiIdentifier string, environments []string, organizationID strin
...
+							}
+							regexRewritePattern := route.GetRoute().GetRegexRewrite().GetPattern().GetRegex()
+							regexRewritePattern = strings.ReplaceAll(
+								regexRewritePattern,
+								newLatestMinorRangeApi.Version,
+								GetMinorVersionRangeRegex(*newLatestMinorRangeApi),
+							)
+							route.Match.PathSpecifier = pathSpecifier
+							action := &routev3.Route_Route{}
+							action = route.Action.(*routev3.Route_Route)
```
slahirucd7, line 904: Can we improve considering below structure ?
```
action := &routev3.Route_Route{
		Action: &routev3.RouteAction{
....
```

adapter/internal/discovery/xds/server_utils.go
chathuranga95: Since the `Patch` attribute is declared as an optional attribute, we have to use with `*`

adapter/internal/discovery/xds/server.go
chathuranga95: This is not possible since the full object construction is not possible using the data we have on the current point of the execution.

adapter/internal/discovery/xds/semanticVersioning.go
```diff
@@ -0,0 +1,404 @@
+/*
```
VirajSalaka, line 1: I think the convention we follow is underscore.

adapter/internal/discovery/xds/semanticVersioning.go
```diff
@@ -0,0 +1,404 @@
...
+	}
+	patchVersion := strconv.Itoa(*semVersion.Patch)
+	return "(v" + majorVersion + "(\\." + minorVersion + "(\\." + patchVersion + ")?)?|" + versionRegexString + ")"
+}
+
+// GetMinorVersionRangeRegex generates minor version compatible range regex for the given version
+func GetMinorVersionRangeRegex(semVersion SemVersion) string {
+	versionRegexString := GetVersionMatchRegex(semVersion.Version)
+	if semVersion.Patch == nil {
+		return "(v" + versionRegexString + "|" + versionRegexString + ")"
```
VirajSalaka, line 107: stick to one variable

adapter/internal/discovery/xds/semanticVersioning.go
```diff
@@ -0,0 +1,404 @@
...
+	isLatestMajorVersion := !isMajorRangeRegexAvailable || CompareSemanticVersions(existingMajorRangeLatestSemVersion, *apiSemVersion)
+	isLatestMinorVersion := !isMinorRangeRegexAvailable || CompareSemanticVersions(existingMinorRangeLatestSemVersion, *apiSemVersion)
+
+	// Remove the existing regexes from the path specifier when latest major and/or minor version is available
+	if (isMajorRangeRegexAvailable || isMinorRangeRegexAvailable) && (isLatestMajorVersion || isLatestMinorVersion) {
+		// Organization's all apis
+		for apiUUID, swagger := range orgIDAPIMgwSwaggerMap[organizationID] {
+			// API's all versions in the same vHost
+			if swagger.GetTitle() == apiName && swagger.GetVHost() == vHost {
+				if isMajorRangeRegexAvailable && swagger.GetVersion() == existingMajorRangeLatestSemVersion.Version ||
```
VirajSalaka, line 178: Better to add parantheses to improve readability.

adapter/internal/discovery/xds/semanticVersioning.go
```diff
@@ -0,0 +1,404 @@
...
+							regexRewritePattern = strings.Replace(regexRewritePattern, GetMajorMinorVersionRangeRegex(existingMajorRangeLatestSemVersion), GetMinorVersionRangeRegex(existingMajorRangeLatestSemVersion), 1)
+						}
+						pathSpecifier := &routev3.RouteMatch_SafeRegex{
+							SafeRegex: &envoy_type_matcherv3.RegexMatcher{
+								Regex: regex,
+							},
+						}
+						route.Match.PathSpecifier = pathSpecifier
+						action := &routev3.Route_Route{}
+						action = route.Action.(*routev3.Route_Route)
```
VirajSalaka, line 202:
```suggestion
						action := route.Action.(*routev3.Route_Route)
```

adapter/internal/discovery/xds/semanticVersioning.go
```diff
@@ -0,0 +1,403 @@
...
+	}
+	patchVersion := strconv.Itoa(*semVersion.Patch)
+	return "v" + majorVersion + "(\\." + minorVersion + "(\\." + patchVersion + ")?)?"
+}
+
+// GetMinorVersionRangeRegex generates minor version compatible range regex for the given version
+func GetMinorVersionRangeRegex(semVersion SemVersion) string {
+	versionRegexString := GetVersionMatchRegex(semVersion.Version)
+	if semVersion.Patch == nil {
+		return versionRegexString
```
renuka-fernando, line 106: `versionRegexString` is only used inside if block. Hence, we can move the var assignment inside the if block, or just return
```go
return GetVersionMatchRegex(semVersion.Version)
```

adapter/internal/discovery/xds/semanticVersioning.go
```diff
@@ -0,0 +1,403 @@
...
+				latestVersions[GetMajorVersionRange(*apiSemVersion)] = *apiSemVersion
+				latestVersions[GetMinorVersionRange(*apiSemVersion)] = *apiSemVersion
+				orgIDLatestAPIVersionMap[organizationID][apiRangeIdentifier] = latestVersions
+			} else {
+				if isLatestMajorVersion {
+					orgIDLatestAPIVersionMap[organizationID][apiRangeIdentifier][GetMajorVersionRange(*apiSemVersion)] = *apiSemVersion
+				}
+				orgIDLatestAPIVersionMap[organizationID][apiRangeIdentifier][GetMinorVersionRange(*apiSemVersion)] = *apiSemVersion
+			}
+		}
```
VirajSalaka, line 232:
```suggestion
if _, ok := orgIDLatestAPIVersionMap[organizationID]; !ok {
	apiVersionMap := make(map[string]map[string]SemVersion)
	orgIDLatestAPIVersionMap[organizationID] = apiVersionMap
}
if _, ok := orgIDLatestAPIVersionMap[organizationID][apiRangeIdentifier]; !ok {
	latestVersions := make(map[string]SemVersion)
	orgIDLatestAPIVersionMap[organizationID][apiRangeIdentifier] = latestVersions
}
if isLatestMajorVersion {
	orgIDLatestAPIVersionMap[organizationID][apiRangeIdentifier][GetMajorVersionRange(*apiSemVersion)] = *apiSemVersion
}
orgIDLatestAPIVersionMap[organizationID][apiRangeIdentifier][GetMinorVersionRange(*apiSemVersion)] = *apiSemVersion
```

adapter/internal/discovery/xds/semanticVersioning.go
```diff
@@ -0,0 +1,403 @@
...
+		if deletingAPIsMajorRangeLatestAPISemVersion.Version == api.GetVersion() {
+			newLatestMajorRangeAPI := &SemVersion{
+				Version: "",
+				Major:   deletingAPISemVersion.Major,
+				Minor:   0,
+				Patch:   nil,
+			}
+			for apiUUID, swagger := range orgIDAPIMgwSwaggerMap[organizationID] {
+				// Iterate all the API versions other than the deleting API itself
+				if swagger.GetTitle() == api.GetTitle() && apiUUID != apiIdentifier {
```
VirajSalaka, line 285: apiIdentifier -> vhost:apiUUID

adapter/internal/discovery/xds/semanticVersioning.go
```diff
@@ -0,0 +1,403 @@
...
+								newLatestMajorRangeAPI = currentAPISemVersion
+								newLatestMajorRangeAPIUUID = apiUUID
+							}
+						}
+					}
+				}
+			}
+			if newLatestMajorRangeAPIUUID != "" {
+				orgIDLatestAPIVersionMap[organizationID][apiRangeIdentifier][majorVersionRange] = *newLatestMajorRangeAPI
+				for _, route := range orgIDOpenAPIRoutesMap[organizationID][newLatestMajorRangeAPIUUID] {
```
VirajSalaka, line 299: newLatestMajorRangeAPIUUID needs to be combination of vhost and uuid

adapter/internal/discovery/xds/semanticVersioning.go
chathuranga95: The `apiIdentifier` is already a unique identifier with vhost and apiUUID.

The `range orgIDAPIMgwSwaggerMap[organizationID]` is also providing the same kind of identifier. Hence, renamed `apiUUID` --> `currentApiIdentifier` to avoid confusion

adapter/internal/discovery/xds/semanticVersioning.go
chathuranga95: This is already a combination of vhost and uuid. Renamed the variable to avoid confusion

VirajSalaka (dismissed):

VirajSalaka (approved):

choreo-cicd:
[succeeded] Dataplane(NorthEU) cluster : dev-deployment-v2 : 20230731.17

choreo-cicd:
[succeeded] Controlplane cluster : dev-deployment-v2 : 20230731.17

choreo-cicd:
[succeeded] Dataplane(EastUS) cluster : dev-deployment-v2 : 20230731.17

choreo-cicd:
[succeeded] Controlplane cluster : stage-deployment-v2 : 20230803.2

choreo-cicd:
[succeeded] Dataplane(EastUS) cluster : stage-deployment-v2 : 20230803.2

choreo-cicd:
[succeeded] Dataplane(NorthEU) cluster : stage-deployment-v2 : 20230803.2

choreo-cicd:
[succeeded] Controlplane cluster : prod-deployment-v2 : 20230804.1

choreo-cicd:
[succeeded] Dataplane(NorthEU) cluster : prod-deployment-v2 : 20230804.1

PR #3682, opened 2025-05-28
demo commit

slahirucd7:
### Issues

Fixes #

### Automation tests
 - Unit tests added: Yes/No

### Tested environments

Not Tested

---
- [ ] Assigned 'Type' label
- [ ] Assigned the project

### DIFF
--- a/enforcer-parent/enforcer/src/main/resources/check_health.sh
+++ b/enforcer-parent/enforcer/src/main/resources/check_health.sh
@@ -15,6 +15,8 @@
 # limitations under the License.
 # -----------------------------------------------------------------------
 
+
+
 ENFORCER_PORT="${ENFORCER_PORT:-8081}"
 grpc_health_probe -addr "localhost:${ENFORCER_PORT}" \
     -tls \


adapter/pkg/messaging/aws_active_mq_connection.go
```diff
@@ -0,0 +1,66 @@
...
+	)
+	if err != nil {
+		logger.LoggerMgw.Infof("Failed to connect to AWS ActiveMQ broker: %s" + err.Error())
+		return nil, err
+	}
+	return con, nil
+}
+
+// InitAwsActiveMqReceiverAndValidate initializes a receiver for the specified topic on the AWS ActiveMQ broker
+func InitAwsActiveMqReceiverAndValidate(connectionString string, topicName string, con *amqp.Conn, maximumallowedUnacknowledgedMessages int) (*amqp.Receiver, error) {
```
malinthaprasan, line 48: connectionString seems to be unused. Please refactor later

adapter/pkg/messaging/aws_active_mq_connection.go
```diff
@@ -0,0 +1,66 @@
...
+		connectionString,
+		&amqp.ConnOptions{
+			SASLType: amqp.SASLTypePlain(
+				userName,
+				password),
+			IdleTimeout: idleTimeoutDuration,
+		},
+	)
+	if err != nil {
+		logger.LoggerMgw.Infof("Failed to connect to AWS ActiveMQ broker: %s" + err.Error())
```
malinthaprasan, line 41: This should be an error log? (Errorf)

malinthaprasan: And also we haven't  err.Error() as an argument. Its an string concat. Won't this cause an error at runtime?

malinthaprasan, line 41:
```suggestion
		logger.LoggerMgw.Infof("Failed to connect to AWS ActiveMQ broker. error: %s", err.Error())
```

adapter/pkg/messaging/aws_active_mq_connection.go
```diff
@@ -0,0 +1,66 @@
...
+		connectionString,
+		&amqp.ConnOptions{
+			SASLType: amqp.SASLTypePlain(
+				userName,
+				password),
+			IdleTimeout: idleTimeoutDuration,
+		},
+	)
+	if err != nil {
+		logger.LoggerMgw.Infof("Failed to connect to AWS ActiveMQ broker. error: %s", err.Error())
```
malinthaprasan, line 41: Fix log mode (Errorf)

adapter/pkg/messaging/aws_active_mq_connection.go
```diff
@@ -0,0 +1,66 @@
...
+		return nil, err
+	}
+	return con, nil
+}
+
+// InitAwsActiveMqReceiverAndValidate initializes a receiver for the specified topic on the AWS ActiveMQ broker
+func InitAwsActiveMqReceiverAndValidate(connectionString string, topicName string, con *amqp.Conn, maximumallowedUnacknowledgedMessages int) (*amqp.Receiver, error) {
+	session, sessionErr := con.NewSession(context.Background(), nil)
```
malinthaprasan, line 49: ChatGPT suggested that we should pass the original context into the new session method.
con.NewSession(ctx, nil)

```
Using context.Background() there means you lose any deadline or cancellation from the caller’s ctx. It’s almost always better to propagate the original ctx into NewSession and NewReceiver so cancellations and timeouts work end-to-end (see Conn.NewSession(ctx, opts) signature 
[pkg.go.dev](https://pkg.go.dev/github.com/Azure/go-amqp)
).
```

```
In Go, context.Context is a built-in interface for carrying deadlines, cancellation signals, and request-scoped values across API boundaries and between goroutines. You’ll see it used almost everywhere in modern Go code, especially for I/O, RPCs, HTTP handlers, database calls, etc.
```

```
context.Background()
What it is:
A ready-made, empty root context. It has no deadline, cannot be cancelled, and carries no values.

When to use it:

As the top-level parent for requests that really have no external cancellation or timeout.

In main(), in initialization code, or in tests when you just need a non-nil Context and don’t care about cancellation.
```

```
Why not always use Background()?
If you call library functions with context.Background(), you:

Lose cancellation — callers cannot cancel mid-flight.

Lose timeouts — operations may hang indefinitely on network blips.

Lose request scoping — you can’t carry metadata (user ID, trace ID, etc.) through your call chain.

That’s why in your AWS ActiveMQ example you should propagate the caller’s ctx (which might carry a deadline or cancellation) all the way through to NewSession(ctx, …) and NewReceiver(ctx, …), rather than dropping back to context.Background().
```

adapter/pkg/messaging/aws_active_mq_connection.go
```diff
@@ -0,0 +1,66 @@
...
+	if sessionErr != nil {
+		logger.LoggerMgw.Errorf("Failed to create session for topic: %s error:%s", topicName, sessionErr.Error())
+		return nil, sessionErr
+	}
+	receiver, receiverErr := session.NewReceiver(
+		context.Background(),
```
malinthaprasan, line 55: Same applies here as well.
https://github.com/wso2/product-microgateway/pull/3682/files#r2115115135

adapter/pkg/messaging/aws_active_mq_connection.go
```diff
@@ -0,0 +1,66 @@
...
+ *
+ */
+
+package messaging
+
+import (
+	"context"
+	"time"
+
+	"github.com/Azure/go-amqp"
```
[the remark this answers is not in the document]
slahirucd7: This is not azure specific. This is a GoLang AMQP library written by Azure.

choreo-cicd:
[] Dataplane(NorthEU) cluster : dev-deployment-v2 : 20250530.7

choreo-cicd:
[] Dataplane(EastUS) cluster : dev-deployment-v2 : 20250530.7

